PR #1, opened 2019-05-12
Set up initial package structure

martindurant:
@hdsingh , this is a typical python package structure. You can install from this directory with
```
pip install -e . --no-deps
```
and the package will be importable (`import xrviz`) and changes made in the inner directory will be available to python. We will probably want modules for the different components, in a similar way to intake.gui . As it stands, only `__version__` is defined, which is auto-generated from the current git tag and commits. The requirements.txt file should include things like `panel` and `xarray`, but that can wait until release time.

I also made a tests folder. I normally use `pytest` for testing, as it is simple but powerful, and is the framework used by Intake. If you've never used it, I can show you how.

I am posting this code as a PR, since this is how I would like to see us contribute and discuss code. Feel free to comment on this thread, or on specific lines of code.


hdsingh:
> I also made a tests folder. I normally use pytest for testing, as it is simple but powerful, and is the framework used by Intake. If you've never used it, I can show you how.

@martindurant I haven't used `pytest`  so I will look into [Pytest Docs](https://docs.pytest.org) but if there is more or some specific things please let me know.

I will also go through `intake/jsignell/panel_gui` branch to understand how to structure code and modules.

Thanks!

martindurant:
Al of the code in that branch has been merged into master already, in the directory https://github.com/intake/intake/tree/master/intake/gui (intake.gui). It is not the only way to do things, you have some freedom in the style that you prefer, but I definitely want things to be as modular, self-contained and reusable as possible.

hdsingh:
Yes sir, I will keep these points in mind while coding and arranging.

PR #2, opened 2019-06-01
Display and description section

hdsingh:
The entire section could be seen by running `python control_section.py`.

I have used jinja template for dynamically  generating HTML. This way HTML/CSS and code could be kept sepatate. As of now I haven't done much styling of the templates. 

I have arranged the modules to have syntax as:
`from xrviz.module.section import Module`, 
for example: 
`from xrviz.control.section import Control`, 
`from xrviz.display.section import Display`


martindurant:
Thanks @hdsingh , I will look in detail as soon as I can.

> I have used jinja template for dynamically generating HTML. This way HTML/CSS and code could be kept separate

This definitely seems like a good idea

You will notice that the version of SigSlot in [dfviz](https://github.com/martindurant/dfviz/blob/master/dfviz/widget.py#L27) has gained some more docstrings since the version you have used. We should probably agree where the canonical version of this class should live. Eventually that place should be Panel, but that will take some time.

The current file structure looks as follows:
```
xrviz/_version.py
xrviz/sigslot/__init__.py
xrviz/sigslot/base.py
xrviz/tests/__init__.py
xrviz/__init__.py
xrviz/describe/__init__.py
xrviz/describe/section.py
xrviz/display/__init__.py
xrviz/display/section.py
xrviz/control/__init__.py
xrviz/control/section.py
```
I would warn against having multiple modules with the same name ("section") and think this is probably over-engineered for our purposes: you could have had `xrviz/describe.py`, etc.

hdsingh:
Yes sir, I will rearrange this. Shall I arrange it as follows?

```
  xrviz/__init__.py
  xrviz/_version.py
  xrviz/tests/__init__.py
  xrviz/sigslot.py
  xrviz/describe.py
  xrviz/display.py
  xrviz/control.py
  xrviz/other_section.py
```

martindurant:
Yes, i believe that would be best.

On June 1, 2019 10:31:42 AM EDT, Harman Deep Singh <notifications@github.com> wrote:
>Yes sir, I will rearrange this. Shall I arrange it as follows?
>
>```
>  xrviz/__init__.py
>  xrviz/_version.py
>  xrviz/tests/__init__.py
>  xrviz/sigslot.py
>  xrviz/describe.py
>  xrviz/display.py
>  xrviz/control.py
>  xrviz/other_section.py
>```
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/intake/xrviz/pull/2#issuecomment-497949657

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

martindurant:
General comment on the code: you will want to turn on flake8/pep8 format checking in your IDE. I don't think there's a need to adhere to the rules too strictly at this point, but they should be kept as a general guide.

I have had a quick chance to run and play with the two example files, and have some initial thoughts.
I think I may have been a little too brief in describing what I think the parts of the interface ought to have. A good comparison is the text view of the xarray instance:
```
Dimensions:    (nx: 61, ny: 25, sigma: 20, time: 60)
Coordinates:
  * time       (time) datetime64[ns] 2019-03-20T06:59:45.937500032 ... 2019-03-22T18:00:00
  * sigma      (sigma) float32 0.0 0.0227 0.0454 0.0681 ... 0.8853 0.9534 1.0
Dimensions without coordinates: nx, ny
Data variables:
    validtime  (time) |S19 ...
    lon        (ny, nx) float32 ...
    lat        (ny, nx) float32 ...
    mask       (ny, nx) float32 ...
    depth      (ny, nx) float32 ...
    zeta       (time, ny, nx) float32 ...
    air_u      (time, ny, nx) float32 ...
    air_v      (time, ny, nx) float32 ...
    u          (time, sigma, ny, nx) float32 ...
    v          (time, sigma, ny, nx) float32 ...
    temp       (time, sigma, ny, nx) float32 ...
Attributes:
    datum1:         reference to LWD=zeta
    datum2:         reference to IGLD85=zeta - IGLD85
    file_type:      Full_Grid
    Conventions:    COARDS
    grid_type:      rectilinear
    z_type:         2-D
    model:          POMGL
    title:          ofs4:/comf/operations/ohms
    comment:        Linux
    source:         current,zeta,temp,wind
    institution:    NOAA/NOS/CO-OPS
    history:        Operation-CO-OPS
    references:     greg.mott@noaa.gov
    creation_date:  2019-03-20  6:51:03  00:00
```
and for one of the variables
```
<xarray.DataArray 'u' (time: 60, sigma: 20, ny: 25, nx: 61)>
[1830000 values with dtype=float32]
Coordinates:
  * time     (time) datetime64[ns] 2019-03-20T06:59:45.937500032 ... 2019-03-22T18:00:00
  * sigma    (sigma) float32 0.0 0.0227 0.0454 0.0681 ... 0.8853 0.9534 1.0
Dimensions without coordinates: ny, nx
Attributes:
    long_name:      Eastward Water Velocity
    units:          m/s
    standard_name:  eastward_sea_water_velocity
```

- the main selection for the user is between *variables*; particularly those which are not also dimensions. Since every coordinate is also a dimension and a variable, it makes no sense to include these in the selector. Variables themselves can be structures in a hierarchical tree, but not for this dataset
- attributes are not selectable things, they should just be displayed: those that belong to the whole dataset should always be visible, plus any that belong to the currently selected variable
- the most important attributes of a variable (see the text version) are the name, data type, coordinates that describe its axes and total size in number of points and bytes.
- in this particular case, and @rsignell-usgs will give more details on this, there is a mapping between the dimensionless nx,ny and variables lat,lon, such that a variable like `u  (time, sigma, ny, nx) float32` can be viewed as `u (time, sigma, lat, lon) float32`.

hdsingh:
@rsignell-usgs @martindurant  Please have a look.

Screenshot from 2019-06-02 22-29-21

Screenshot from 2019-06-02 22-29-26
Screenshot from 2019-06-02 22-29-43
Screenshot from 2019-06-02 22-29-50

I think it is better to just show all the four basic properties in the selector :
 `Dimensions`,`Coordinates`,`Variables`, `Attributes`, since displaying them all together would require more space and not look attractive. But now only the `Varaibles` expand upon selection.
The plot button would be disabled for the un-plotable properties (dims,coords,attrs) and even some variables like `validtime` in this case.

hdsingh:
> Since every coordinate is also a dimension and a variable, it makes no sense to include these in the selector.

@rsignell-usgs @martindurant  I understand this fact however there are other properties associated with each of `Coords` like `name`,`shape`,`n_bytes`,`dtype`,`units` and `attributes`. If we don't add `Coord`  in displayer, these properties would not be displayed. 
Shall I design in such a way that on clicking a coord, its properties would be displayed similar to variables? However the main difference would be that `Create Plot` button is disabled for coords.

rsignell-usgs:
@hdsingh , I like the idea of drilling into the `Coords` to see the properties -- I think it will provide a more consistent feel to the interface, and I think users would expect it.

hdsingh:
So this is how `Coordinates` appear now:

Screenshot from 2019-06-03 18-14-21

martindurant:
I disagree! `time` also appears in the variables section, so you have redundant information - indeed, there is no need for the expanding tree structure at all.
```
>>> ds = xr.open_dataset('great_lakes.nc')
>>> list(ds.variables)
['validtime',
 'time',
 'lon',
 'lat',
 'mask',
 'depth',
 'sigma',
 'zeta',
 'air_u',
 'air_v',
 'u',
 'v',
 'temp']
```

martindurant:
For discussion: https://github.com/martindurant/xrviz/tree/simpler/xrviz

Screen Shot 2019-06-03 at 10 30 27

hdsingh:
@martindurant I ran and used the simpler interface, here we are only displaying `Variables`. This simpler interface also looks good and amazing!

> I disagree! time also appears in the variables section, so you have redundant information - indeed, there is no need for the expanding tree structure at all.

We are using `list(ds.data_vars)` instead of `list(ds.variables)` for the variables section so I think there is not much chance of repetetion. That's why `time` is not present in both simpler and this interface. (I also tested this for these [two other datasets](https://github.com/hdsingh/Dashboards/blob/master/data/)).

I think that our aim is to get to know our xarray dataSet/Array as much as possible without coding, that's why I listed all the properties, to replace the text view of the xarray instance.

However, I will change as you both say and agree upon.
@rsignell-usgs Please let me know the final plan.

hdsingh:
This is how the interface looks for `DataArray`:

Screenshot from 2019-06-04 00-06-37

martindurant:
Ah, but if you have a DataArray, there is nothing to select, the selection is already made.

hdsingh:
Oh! yes!! I will correct this. Thanks!

hdsingh:
Now selection and description of DataArray happens upon initialisation.

hdsingh:
@martindurant  I have implemented `set_dataset_property` method to pre-set the selections of dataset properties. 

It could be used as follows:
``` python
control = Control(ds)
control.set_dataset_property('Variables')
```
If this looks fine shall I set other methods `set_variable`, `set_coordinate` too  to pre-set a particular `data_var` or `coord` ?

martindurant:
I think we should talk about this before you implement

rsignell-usgs:
Yes, coordinates are just variables, but xarray can distinguish between coordinates and data variables with the  `ds.coords` and `ds.data_vars` methods, and I think they should be handled separately if they exist.

hdsingh:
@martindurant I have designed a simple test for the `Display.fill_items` to work on running `pytest` in outer xrviz/ directory, based on example test.

martindurant:
@hdsingh , my version with the following change, as we discussed 
```diff
--- a/xrviz/display.py
+++ b/xrviz/display.py
@@ -38,6 +38,15 @@ class Display(SigSlot):
         self.data = data
         self.select = pn.widgets.MultiSelect(size=8, min_width=300,
                                              width_policy='min',
-                                             options=list(data))
+                                             options={
+                                                 name+('\U0001f30d' if isgeo(data[name])
+                                                 else ""): name
+                                                 for name in data.variables}
+                                             )
         self._register(self.select, "variable_selected")
         self.panel = pn.Row(self.select)
+
+
+def isgeo(var):
+    return getattr(var, 'standard_name', "").lower() in [
+        'latitude', 'longitude', 'time']
```

Screen Shot 2019-06-05 at 11 00 36

I am not convinced this is the *right* styling, and if may indeed require metpy to know which fields pass the test (I am only guessing).

hdsingh:
@rsignell-usgs @martindurant According to discussion and example above, as of now, `is_geo()` is  temporarily using `standard_name` to determine whether a variable is part of CF Convention or not. Going through [CF coordinate-types](http://cfconventions.org/cf-conventions/cf-conventions.html#coordinate-types), I realise we should use `units` to determine, since `standard_name` is optional. Shall I change `is_geo` according to this?

martindurant:
I was only guessing about how one might determine is a coordinate is geo/CF or not. Certainly use the canonical definition, if you have it. Probably more important than annotating geo is annotating variables which are coordinates, perhaps with "📈"

hdsingh:
I have implemented `set_variables` and `select_variable` method. The following program illustrates its usage.

```
import xrviz as xv
from xrviz.control import Control
import panel as pn
pn.extension()

ds = xv.sample_data.great_lakes
dashboard = Control(ds)
dashboard.displayer.set_variables(['lon','zeta','air_v','air_u'])
dashboard.displayer.select_variable('air_u')
dashboard.panel
```
`set_data` needs some more improvement now.

martindurant:
This code looks like a test :)

I am a little surprised by `.set_variables` - are you wishing to provide an override, so the user can choose to display only a subset of the variables? I'm not sure it is necessary to give this flexibility.
(and I think you can then use `select_variable` to choose something that's not in the list)

martindurant:
I think it would be easier to remove [this branch](https://github.com/intake/xrviz/pull/2/commits/2bbf6e3fb961dc3434fd038a3dd0a081c9a2bc25#diff-24ce523b44653b3783546ac0237671f9R65) for now. Let's keep things simple, since the next panel with the field options is much more complicated.

martindurant:
> Test and fields panel are next in line

"Test" is the output? I would suggest the basic fields panel (x axis, y axis choices to start with) and output panel, so that we can start having some examples to show. But yes, that is in the next PR, not this one.

hdsingh:
> "Test" is the output?

I mean the tests for Display and Description Section.

martindurant:
OK - no need to make too many tests

hdsingh:
These are the units I found in [NetCDF Climate and Forecast (CF) Metadata Conventions](http://cfconventions.org/cf-conventions/cf-conventions.html#coordinate-types).

Latitude Coordinate: `'degrees_north','degree_north','degree_N,'degrees_N','degreeN','degreesN'`
Longitude Coordinate: `'degrees_east','degree_east', 'degree_E', 'degrees_E', 'degreeE', 'degreesE'`
Vertical (Height or Depth) Coordinate: `'meters'` 
Dimensional Vertical Coordinate: `bar, millibar, decibar, atmosphere (atm), pascal (Pa), and hPa`  
Dimensionless Vertical Coordinate: `level, layer, or sigma_level`
Time Coordinate: `day (d), hour (hr, h), minute (min) and second (sec, s)`
 
 These cover almost every variable. So having label only for coords, as you have stated earlier is better idea. 
 Example: For this particular dataset(Greatlakes) only `temp`'s units are not present in these units.
 So shall we only annotating variables which are coordinates?

PR #4, opened 2019-06-15
Aggregations

hdsingh:
I have implemented these options:
- max
- min
- mean
- stddev
- median
- count (of non-nan)

Each of the option supports aggregation along multiple dims.

Screenshot from 2019-06-15 22-13-35

@rsignell-usgs @martindurant Please have a look.

### DIFF
--- a/xrviz/fields.py
+++ b/xrviz/fields.py
@@ -24,7 +24,7 @@ class Fields(SigSlot):
 
     def __init__(self, data):
         super().__init__()
-        self.data = data
+        self.set_data(data)
         self.x = pn.widgets.Select(name='x', width=200)
         self.y = pn.widgets.Select(name='y', width=200)
         self.agg_selectors = pn.Column()
@@ -48,8 +48,16 @@ class Fields(SigSlot):
                             self.agg_graph,
                             name='Fields',)
 
-        if isinstance(data, xr.DataArray):
+        if not self.is_dataset:
             self.setup(data)
+
+    def set_data(self, data):
+        if isinstance(data, xr.Dataset):
+            self.data = data
+            self.is_dataset = True
+        else:
+            self.data = data
+            self.is_dataset = False
 
     def setup(self, var):
         # print(self.panel)
@@ -60,7 +68,7 @@ class Fields(SigSlot):
         self.agg_plot_button.disabled = True
         self.agg_graph[0] = pn.Spacer(name='Agg Graph')  # To clear Agg Graph upon selection of new variable
 
-        if isinstance(self.data, xr.Dataset):
+        if self.is_dataset:
             if isinstance(var, str):
                 self.var = var
             else:
@@ -110,7 +118,10 @@ class Fields(SigSlot):
 
     def create_agg_plot(self, *args):
         sel_dims = self.dims_selected_for_agg
-        sel = getattr(self.data, self.var)
+        if self.is_dataset:
+            sel = getattr(self.data, self.var)
+        else:
+            sel = self.data
 
         for dim, agg in sel_dims.items():
             if agg == 'count':
--- a/xrviz/fields.py
+++ b/xrviz/fields.py
@@ -31,21 +31,18 @@ class Fields(SigSlot):
         self.agg_opts = ['None', 'mean', 'max',
                          'min', 'median',
                          'std', 'count']
-        self.agg_plot_button = pn.widgets.Button(name='Aggregate',
-                                                 width=100,
-                                                 disabled=True)
+
         self.agg_graph = pn.Row(pn.Spacer(name='Agg Graph'))
 
         self._register(self.x, 'x')
-        self._register(self.agg_plot_button, 'agg_plot_button', 'clicks')
+        self._register(self.y, 'y')
 
         self.connect('x', self.change_y)
-        self.connect('agg_plot_button', self.create_agg_plot)
+        self.connect('y', self.change_dim_selectors)
 
         self.panel = pn.Row(pn.Column(self.x, self.y),
                             pn.Column(self.agg_selectors,
                                       pn.Spacer(name='agg_plot_button')),
-                            self.agg_graph,
                             name='Fields',)
 
         if not self.is_dataset:
@@ -61,8 +58,6 @@ class Fields(SigSlot):
 
     def setup(self, var):
         self.agg_selectors.clear()  # To empty previouly selected value from selector
-        self.panel[1][1] = self.agg_plot_button
-        self.agg_plot_button.disabled = True
         self.agg_graph[0] = pn.Spacer(name='Agg Graph')  # To clear Agg Graph upon selection of new variable
 
         if self.is_dataset:
@@ -76,14 +71,6 @@ class Fields(SigSlot):
         #  dims_aggs: for ex {'dim1':'None','dim2':'None'}
         self.dims_aggs = dict(zip(self.var_dims, ['None']*len(self.var_dims)))
 
-        for dim in self.var_dims:
-            dim_selector = pn.widgets.Select(name=dim,
-                                             options=self.agg_opts,
-                                             width=100,)
-            dim_selector.param.watch(self.callback, ['value'],
-                                     onlychanged=False)
-            self.agg_selectors.append(dim_selector)
-
         x_opts = self.var_dims.copy()
         if len(x_opts) > 0:  # to check that data has dim (is not Empty)
             self.x.options = x_opts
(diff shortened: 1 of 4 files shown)


xrviz/fields.py
```diff
@@ -64,6 +99,48 @@ def change_y(self, value):
         values.remove(self.x.value)
         self.y.options = values
 
+    def create_agg_plot(self, *args):
+        remaining_dims = [dim for dim in self.var_dims if dim not in self.dims_selected_for_agg]
+        assign_opts = {dim: self.data[dim] for dim in remaining_dims}
+
+        # the method getattr(data.air_u, self.aggregation.value ,[*self.dims_selected_for_agg])
+        # does not works as expected
+        if self.aggregation.value == 'mean':
```
hdsingh: Each statement has been written separately  because the method `getattr(data.air_u, self.aggregation.value ,[*self.dims_selected_for_agg])` does not works as expected. 
For example data: `getattr(data.air_u, 'mean' ,['time'])()` returns 
```
<xarray.DataArray 'air_u' ()>
array(1.03431, dtype=float32)
```
which is not what we want. Please let me know if there is correct way to do this. Else I will try raising an issue.

hdsingh:
@martindurant  @rsignell-usgs All the options have been implemented so this PR can be reviewed now.

xrviz/fields.py
martindurant:
```
getattr(data.air_u, 'mean')(['time'])
```

hdsingh: I have changed according to this. Thanks!

martindurant:
You will need to have the option of separate aggregations for each of the extra dimensions: you could have mean along time and count along sig. Why the separate "aggregate" button, doesn't this happen at the same time as generating the plot?

Good start, though.

martindurant:
Furthermore, dimensions selected for x and y should not appear in the dimensions available for aggregation, and any dimension *not* aggregated should be available for selection as before (or you could have an aggregation called "selection" in the list).

hdsingh:
> You will need to have the option of separate aggregations for each of the extra dimensions: you could have mean along time and count along sig.

I haven't been able to understand what exactly you mean by this point. We already are able to find mean along time and count along sigma for any of the variable having these as dimensions.

> Why the separate "aggregate" button, doesn't this happen at the same time as generating the plot?

This point seems valid as long as we are looking only at this PR. But in the `set_reset_coords` branch, we are having both default and custom coords along with the dimensions, but the `Aggregate along` multiselect is only displaying the dims. The case where `x` and `y` show coords, having a same button is not applicable. We can't aggregate along coords, so we need to have this button. To have a look at both the `aggregation` and `set_reset_coords` work together, please see [temp_set_reset](https://github.com/hdsingh/xrviz/tree/temp_set_reset) branch. 

@rsignell-usgs Please let me know your views on this.

martindurant:
> We already are able to find mean along time and count along sigma

In your interface, this does not appear to be the case, you only have one aggregation function, which is applied to all selected dimensions.

> both default and custom coords along with the dimensions

@rsignell-usgs , is it possible to aggregate over a mapped coordinate, such as `.mean("lat")` for the example dataset?

hdsingh:
I tried `data.temp.mean(['lat'])` and got 
```
ValueError: 'lat' not found in array dimensions ('time', 'sigma', 'ny', 'nx')
```
So we can only use the dimensions present in variable for aggregation.

martindurant:
Specifically:
```
>>> data.set_coords(['lat', 'lon']).mean(['lat'])
ValueError: Dataset does not contain the dimensions: ['lat']
```

but *can* it be done?

hdsingh:
> but can it be done?

I have tried setting coords but still don't know how it can be done since `mean` or other aggregation is being done over dims of a variable. 

> In your interface, this does not appear to be the case, you only have one aggregation function, which is applied to all selected dimensions.

Do you mean to a have functionality to do  something like `data.temp.mean(['time']).count(['sigma']).hvplot.quadmesh()`.

What image do you have in mind for implementing this? Do we need to have `Aggregate along` selector for each of the dim?

martindurant:
> but can it be done?

My question is addressed to @rsignell-usgs , who maybe can ask other xarray people if necessary.

> Do we need to have Aggregate along selector for each of the dim

exactly

rsignell-usgs:
Aggregations like `.mean()` only make sense when the coordinates are 1D and uniform.   For 2D coordinates like `lon` and `lat` in the Lake Ontario example, it would make sense to take the mean over the entire lake (e.g. the average lake temperature), but you would need a custom function that calculated the approximate area of each cell and weighted the result accordingly.    I think there is limited use for this aggregation capability in the GUI, actually.    I'm not sure it should even be included, on the basis that the additional complexity is not worth the marginal benefit for a few cases.

martindurant:
> I'm not sure it should even be included

OK - just thought there might be a simple way. Perhaps years from now, xrviz may become something like imagej, with all manner of extra functionality plugins ...

hdsingh:
Screenshot from 2019-06-18 13-49-35

Now we can apply any aggregation function to any dimension.

martindurant:
You have selected nx and ny for x and y axes, but you only got one axis in the plot, for time. 
- an axis can be aggregated of chosen for plotting, not both
- axes that are not aggregated (time, in this case) should become selectors, as before.
- as before, I don't understand the purpose of the aggregate button, compared to the previous plot button

(the values of temp shown seem pretty strange)

hdsingh:
Screenshot from 2019-06-18 17-12-49

Only one plot button would be used. Creating Aggregation graph yet needs to be done. Am I going right now?

martindurant:
Yes, this appears to have the right behaviour.

> Creating Aggregation graph yet needs to be done

You mean that the selected option has no effect yet, right?

hdsingh:
> You mean that the selected option has no effect yet, right?

Yes sir, I am working on it.

martindurant:
We can discuss this tomorrow, but what do you think ( @rsignell-usgs ) that "None" for the aggregation selection should be replaced by "Select" and "Animate", rather than having both controls show up? That would greatly simplify the output area.

I would like the see labels or boxes around the x/y selectors ("plot dimensions") and the aggregation selectors ("aggregations") in the interface, just to make things clearer for the user.

hdsingh:
The select options are now working for aggregation. Where should this aggregation plot be displayed?

Screenshot from 2019-06-18 20-25-31

(I haven't positioned it yet, so it is on the right side which doesn't looks good)

martindurant:
There should only be one plot, I'm not sure what you are showing here.

hdsingh:
This is simple version of the code used  in creating the two graphs as shown above.
For the case where
`x`:`nx`
`y`:`ny`
`sigma`:`mean`
`time`:`None` (could be any other aggregation)

For 1st graph:
```python
data = xrviz.sample_data.great_lakes
var = 'temp'
sel_data = data[var].isel(sigma=0,time=0)
sel_data = sel_data.assign_coords(nx=data.nx,ny=data.ny)
sel_data.hvplot.quadmesh(x='nx',y='ny')
```

For creating Aggregation graph: ( where only `sigma`'s mean, has been taken.
For any other dimension aggregation would have been done similarly)

```python
sel_data = data[var]
sel_data = getattr(sel_data,'mean')('sigma')
# sel_data = getattr(sel_data,'max')('time') # -->if time's selection had not been None
sel_data = sel_data.assign_coords(time=data.time,
                                  nx= data.nx,
                                  ny=data.ny)
sel_data.hvplot.quadmesh()
```

Please let me know what exactly the graph should be. How should we combine 
the two graphs above to generate what we require?

martindurant:
Since the user in your example has selected mean for aggregation on sigma, I would expect to see only the second graph. If they wanted to select on sigma, they would have chosen None for sigma (or one of "select", "animate", as suggested in my previous comment).

hdsingh:
Screenshot from 2019-06-19 10-15-42

@rsignell-usgs @martindurant  Does this looks fine?

rsignell-usgs:
I'm super confused by this aggregation stuff.  Looking forward to discussing at today's meeting.

martindurant:
> Does this looks fine?

Not quite what I was expecting - we will discuss

rsignell-usgs:
This is all working quite nicely now, IMHO.     @martindurant okay to merge this?

hdsingh:
According to me work here is complete. This could be merged now. Please let me know if there is anything else to change.

rsignell-usgs:
> According to me
I love this phrase.   My Italian colleagues used this often during my time working for NATO, and I miss it!

PR #7, opened 2019-06-17
Set reset coords

hdsingh:
@rsignell-usgs @martindurant I have implemented set/reset coords and plot generation according to our discussion [here](https://github.com/intake/intake-gsoc-gui/issues/20) . There were few points, which I did a bit differently. I would soon add a detailed description of the functionality implemented in this dashboard. In the mean time, please have a look at the interface. 

Screenshot from 2019-06-17 12-20-07


hdsingh:
In case of DataArray, since there is only single variable, `Set coord` has not been implemented. For a `DataSet`, it has been implemented.
In the `Set Coords` tab, Cross selector displays all the variables preset in the DataSet. User selects the variable from left side and move them to right side by `>>` button. After making selection, click on the `Set Coords` button, which  converts the vars to coords. Reverse operation (coord to var) is also possible.
We cannot reset the `indexed_coords`.

Each time user clicks on the `Set Coord`:
- Display section is updated to annotate a coord with ‘📈’ 
- Description section updates.
- Selectors in Fields panel update

For each of `dashboard`,`control`,`display`,`describe`,`fields`,`coord_setter`, a method `set_coords` has been implemented which performs necessary updates required in that particular panel.

In the fields panel, `x` displays all the options (`var_dims+non_indexed_coords`), while `y` is selective. 
- If `var_dim` has been selected in `x`, `y` also displays `var_dim`.
- Same is the case for non_indexed_coords.
- In case of `non_indexed_coords`, dimensionality of two vars need to be same, so only the coords having same `ndim` as `x` are displayed in `y`.

If `dims` have been selected, `data.var.sel` method is used for plotting. 
If `coord` is selected, `data.hvplot.quadmesh(x = 'coord1', y='coord2')` method has been used.

Please have a look by playing around with the interface.

xrviz/dashboard.py
```diff
@@ -57,42 +62,65 @@ def set_data(self, data):
...
         self.output[2].clear()  # clears Players
 
         if self.is_dataset:
-            for dim in self.var_selector_dims:
-                selector = pn.widgets.Select(name=dim, options=list(self.data[self.var][dim].values))
-                self.index_selectors.append(selector)
-                selector.param.watch(self.callback_for_indexed_graph, ['value'], onlychanged=False)
-        else:
+            x = self.kwargs['x']
+            if not self.is_non_indexed_coord(x):  # i.e is a var_dim
```
hdsingh: This case `# i.e is a var_dim` and `# if is_dataArray`, are almost same, except a few lines in between. Initially they were combined (for the lines which are same). Now I have separated them, to accomplish the task here. Please let me know if there is a better way of arranging these three cases here.

hdsingh:
@martindurant @rsignell-usgs Please review the changes.

xrviz/dashboard.py
martindurant: I suspect indeed this could be cleaner, or simply moved into (testable) functions in utils. However, functionality trumps code cleanliness, so let's first ensure that the behaviour is as required.

hdsingh:
Screenshot from 2019-06-21 23-52-37

@rsignell-usgs @martindurant Unlike `dims`, in case of creating graph for a variable with respect to a `coord`, `Aggregations` would be empty. Also there doesn't seem any generalisable way to  know beforehand, along what coords/vars index selectors would be generated. So is it fine if we are displaying both selector and player in the graph output?

rsignell-usgs:
@hdsingh, I was just about to comment on this.   If coords like `lon` and `lat` are selected, it seems that aggregation along other dimensions like `sigma` and `time` should still work.

hdsingh:
Yes,  it seems that we can have aggregation for a variable along a dimension, even when `x` and `y` Select has `coord` as options. I will check and implement this. But in this case Aggregation options will not include `Select` and `Animate` as option because we don't know beforehand for sure, if that dim would be present in the generated index selectors.

hdsingh:
Screenshot from 2019-06-22 11-16-13

We can have aggregation for a variable along a dimension, when x and y Select has `coord` as options.

When a coord has been selected in `x` and `y`, we can only aggregate it along the dims which are not present in `x` or `y`. For example : 
Var: `temp`
x: `lat`
y: `lon`

So now we can't aggregate along dims `nx` and `ny` since they are already present in `lat` and `lon`. Only aggregation only along remaining dims: `time` and `sigma` is possible.

hdsingh:
@rsignell-usgs @martindurant Everything is working as expected. Please have a look.

xrviz/dashboard.py
```diff
@@ -46,72 +51,88 @@ def __init__(self, data):
...
+
+                for dim in self.var_selector_dims:
+                    ops = list(self.data[self.var][dim].values)
+
+                    if self.kwargs[dim] == 'Select':
+                        selector = pn.widgets.Select(name=dim, options=ops)
+                    else:
+                        selector = pn.widgets.DiscretePlayer(name=dim, value=ops[0], options=ops)
+                    self.index_selectors.append(selector)
+                    self._register(selector, selector.name)
```
martindurant: This clears out previous watchers and callbacks for a widget of the same name? It might be nice to clear all events before starting to add new ones.

xrviz/dashboard.py
```diff
@@ -145,4 +166,75 @@ def create_indexed_graph(self):
...
+
+    def is_non_indexed_coord(self, x):
+        indexed_coords = set(self.data[self.var].dims) & set(self.data[self.var].coords)
+        non_indexed_coords = list(set(self.data[self.var].coords) - indexed_coords)
+        if x in non_indexed_coords:
+            return True
+        else:
+            return False
+
+    def set_data(self, data):
```
martindurant:
```
def set_data(self, data):
    self.data = data
    self.is_dataset = isinstance(data, xr.DataSet)
```

Not sure if you want the validation here at all, but if yes, it should give a more informative error; I would prefer it in Dashboard's init, perhaps like
```
if not isinstance(data, xarray.core.dataarray.DataWithCoords):
    raise ValueError("Input should be an xarray data object, not %s" % type(data))
```

xrviz/dashboard.py
```diff
@@ -145,4 +166,75 @@ def create_indexed_graph(self):
...
+            self.data = data
+            self.is_dataset = False
+        else:
+            raise ValueError
+
+    def set_coords(self, *args):
+        # We can't reset indexed coordinates so add them every time
+        # in coord_selector.value
+        self.data = self.data.reset_coords()
+        indexed_coords = set(self.data.dims) & set(self.data.coords)
```
martindurant: Maybe cleaner to call `set.intersection`

xrviz/dashboard.py
```diff
@@ -145,4 +166,75 @@ def create_indexed_graph(self):
...
+        new_coords = list(set(self.control.coord_setter.coord_selector.value) | set(indexed_coords))
```
martindurant: Maybe cleaner to call `set.union`

xrviz/dashboard.py
```diff
@@ -145,4 +166,75 @@ def create_indexed_graph(self):
...
+        self.control.set_coords(self.data)
+
+    def check_is_plottable(self, var):
+        """
+        If a variable is 1-d, disable plot_button for it.
+        """
+        self.plot_button.disabled = False  # important to enable button once disabled
+        if self.is_dataset:
+            var = var[0]
+            if len(list(self.data[var].dims)) <= 1:
```
martindurant: Simpler (and this happens elsewhere too):
```
def check_is_plottable(self, var):
    data = self.data[var] if self.is_dataset else self.data
    self.plot_button.disabled = len(data.dims) <= 1
```

xrviz/describe.py
```diff
@@ -38,7 +38,7 @@ def setup(self, var):
         if isinstance(self.data, xr.Dataset):
             self.panel.object = self.variable_pane_for_dataset(var)
         else:
-            self.panel.object = self.variable_pane_for_dataarray(var)
+            self.panel.object = self.variable_pane_for_dataarray()
```
martindurant: Same pattern of many isinstances here.

I'll ask the question again: can we not view a DataArray as a Dataset which happens to have only one variable? It would simplify a lot of this sort of thing.

xrviz/fields.py
```diff
@@ -60,20 +58,21 @@ def set_data(self, data):
...
         self.agg_selectors.clear()  # To empty previouly selected value from selector
-        self.agg_graph[0] = pn.Spacer(name='Agg Graph')  # To clear Agg Graph upon selection of new variable
 
         if self.is_dataset:
             if isinstance(var, str):
                 self.var = var
             else:
                 self.var = var[0]
             self.var_dims = list(self.data[var].dims)
+            self.indexed_coords = set(self.var_dims) & set(self.data[var].coords)
```
martindurant: This appears to be repeated code.

Same comments above about using set methods rather than operators.
Also, is it weird that the first one remains a set , but the other two are lists?

xrviz/fields.py
```diff
@@ -85,23 +84,46 @@ def setup(self, var):
...
-        self.y.options = values
+         # if x belong to var_dims replace the y with remaining var_dims
+        # else if x belong to non_indexed_coords, replace y with remaining
+        # non_indexed_coords
+        values = self.sel_options.copy()
+        x_val = self.x.value
+        values.remove(x_val)
+        if isinstance(self.data, xr.Dataset):
+            if x_val in self.var_dims:
+                valid_values = list(set(values) - set(self.non_indexed_coords))
```
martindurant: Some more list->set->list with operators; perhaps we need some utils helper functions.

xrviz/fields.py
hdsingh: I have replaced operators with set methods. Since `pn.widget.Select` accepts only `list` or `dict` as options, `self.sel_options` has to be a list and not set.

martindurant: The result of the `list(set(...))` has arbitrary ordering, you might want `sorted(set(...))` instead.

xrviz/dashboard.py
hdsingh: Corrected. Thanks!

xrviz/fields.py
hdsingh: Sorted has been applied to end result, i.e. just before passing as options to x and y.

xrviz/describe.py
hdsingh: I have thought about this. I don't see a way by which this could be done.

`xr.Dataset` and `xr.DataArray` have to be handled separately. As of now issue of code repeatability is there in dashboard.py, which would be resolved after implemetations of other panels. For most of the functions, the part where these two differ are `data[var]` and `data`.

martindurant: The following is a simple way.

In init of any object taking xarray/dataset:
```
if isinstance(data, xr.DataArray):
    data = xr.Dataset({'data': data})
```

xrviz/dashboard.py
hdsingh: Since we can't access the old widgets, once plot button is clicked, it would be better to simply delete those. I have implemented this deletion of old widgets.

xrviz/dashboard.py
hdsingh: Corrected. Thanks!

xrviz/dashboard.py
martindurant: When you say "delete", you mean removing from the `_sigs` and `_map` dicts? Perhaps Siglot should grow a method for doing this.

xrviz/dashboard.py
hdsingh: Corrected. Thanks!

xrviz/dashboard.py
hdsingh: Corrected. Thanks!

xrviz/fields.py
hdsingh: I have simplified this.

xrviz/describe.py
martindurant: or better:
```
data = xr.Dataset({data.name: data}, attrs=data.attrs)
```

hdsingh: Thanks a lot sir!!
This is the simple yet the best part of code. It has really simplified everything.

rsignell-usgs:
With this latest commit, I'm getting:
```
(xrviz) C:\Users\rsignell\Documents\GitHub\xrviz\examples>python control_section.py
Traceback (most recent call last):
  File "control_section.py", line 7, in <module>
    dash = Dashboard(ds)
  File "c:\users\rsignell\documents\github\xrviz\xrviz\dashboard.py", line 34, in __init__
    self.control = Control(self.data)
AttributeError: 'Dashboard' object has no attribute 'data'
```

hdsingh:
Now it is working. I will soon simplify the code using `data = xr.Dataset({data.name: data}, attrs=data.attrs)` as discussed above.

martindurant:
> With this latest commit

I will make a PR, perhaps after Wednesday's meeting, implementing Travis CI. We are probably starting to have enough functionality which is concrete that thorough testing is warranted.

rsignell-usgs:
@hdsingh, when I select `lon` and `lat` as coordinates, I get `x=lat`, `y=lon` by default.   Is it possible to get `x=lon` and `y=lat` as default, or is it impossible to tell what should go where?

hdsingh:
Yes correct defaults need to be set. I will look in detail how this could be done from [this answer](https://stackoverflow.com/questions/53469510/how-to-identify-time-lon-and-lat-coordinates-in-xarray) and will implement this in one of upcoming PRs. As of now options are being displayed according to their alphabetical order.

hdsingh:
Now there is no separate handling for `DataArrays` due to conversion while initialisation by `data = xr.Dataset({data.name: data}, attrs=data.attrs)`.

hdsingh:
@martindurant I have started writing tests. Please have a look and let me know if this looks fine. My aim here is to replicate the user behaviour and check if the interface is functioning as expected.

martindurant:
Yes, these tests look like a good start. I expect they are all passing.
I'll merge this now, so we can discuss next steps tomorrow.

PR #11, opened 2019-06-18
fix pip installation

martindurant:


martindurant:
@rsignell-usgs , this would solve the problem you were seeing. You can test, if you like.

rsignell-usgs:
@martindurant , the environment file that allows me to run the examples is:
```
name: xrviz
channels:
  - conda-forge
  - defaults
dependencies:
  - cartopy
  - hvplot
  - intake
  - intake-xarray
  - nb_conda_kernels
  - netcdf4
  - panel
  - pip
  - xarray
  - pip:
      - git+https://github.com/intake/xrviz
```

martindurant:
I assume this is for your general work - xrviz does not call intake stuff.

Looks like I was probably wrong to assume that listing panel also got hvplot or that listing xarray automatically got the cdf and carto stuff. 

Which package comes from defaults?

rsignell-usgs:
You are right.  Intake is not needed for the current demo.   But everything else is.

martindurant:
OK, so I'll merge this and make a release on conda-forge, just to show we can.

PR #13, opened 2019-06-26
Style panel

hdsingh:
@rsignell-usgs @martindurant This is how the style panel looks:

Screenshot from 2019-06-26 08-32-25

The options available are:
- Height
- Width
- Cmap
- Colorbar(enable or disable)
- Colormap Limits
- Color Scale ( several options such as 'cos, 'square', 'sqrt', 'sin'..)

### DIFF
--- a/xrviz/dashboard.py
+++ b/xrviz/dashboard.py
@@ -44,8 +44,12 @@ class Dashboard(SigSlot):
         self.connect("set_coords", self.set_coords)
 
         self.control.displayer.connect('variable_selected', self.check_is_plottable)
+        self.control.displayer.connect('variable_selected', self.get_selected_data)
+        self.control.displayer.connect('variable_selected', self.link_aggregation_selectors)
+        self.control.fields.connect('x', self.link_aggregation_selectors)
+        self.control.fields.connect('y', self.link_aggregation_selectors)
 
-        self.control.style.connect('colormap_limits', self.change_limits)
+        self.control.style.connect('colormap_limits', self.change_colormap_limits)
 
         self.panel = pn.Column(self.control.panel,
                                self.plot_button,
@@ -54,6 +58,16 @@ class Dashboard(SigSlot):
         # To auto-select in case of single variable
         if len(list(self.data.variables)) == 1:
             self.control.displayer.select.value = list(self.data.variables)
+
+    def link_aggregation_selectors(self, *args):
+        """
+        To link aggregation selectors with color_limits
+        Whenever any aggregation selector changes it value,
+        self.selected_data is updated and the limits are changed.
+        """
+        for dim in self.control.kwargs['remaining_dims']:
+            self.control.fields.connect(dim, self.get_selected_data)
+            self.control.fields.connect(dim, self.change_colormap_limits)
 
     def create_plot(self, *args):
         self.kwargs = self.control.kwargs
@@ -79,7 +93,7 @@ class Dashboard(SigSlot):
             sel_data = self.select_data()
 
             color_range = {sel_data.name: (sel_data.quantile(colormap_limits[0]),
-                                               sel_data.quantile(colormap_limits[1]))}
+                                           sel_data.quantile(colormap_limits[1]))}
             if color_scale is not 'linear':
                 sel_data = getattr(numpy, color_scale)(sel_data)  # Color Scaling
 
@@ -212,6 +226,7 @@ class Dashboard(SigSlot):
         new_coords = set(args[0]).union(indexed_coords)
         self.data = self.data.set_coords(new_coords)  # this `set_coords` belongs to xr.dataset
         self.control.set_coords(self.data)
+        self.link_aggregation_selectors()  # need to link here also since new agg selectors are created
 
     def check_is_plottable(self, var):
         """
@@ -221,11 +236,19 @@ class Dashboard(SigSlot):
         data = self.data[var[0]]
         self.plot_button.disabled = len(data.dims) <= 1
 
-    def change_limits(self, *args):
-        limits = args[0]
+    def change_colormap_limits(self, *args):
+        print(args)
+        print('change_limit')
         sel_data = self.select_data()
+        limits = self.control.style.colormap_limits.value
         c_lim_lower = sel_data.quantile(limits[0])
         c_lim_upper = sel_data.quantile(limits[1])
+        print(str(c_lim_lower.values.round(5)), str(c_lim_upper.values.round(5)))
 
         self.control.style.lower_limit.value = str(c_lim_lower.values.round(5))
         self.control.style.upper_limit.value = str(c_lim_upper.values.round(5))
+
+    def get_selected_data(self, *args):
+        # When a variable changes, we need to find the data,
+        # according to selections made by the user
+        self.selected_data = self.select_data()
--- a/xrviz/dashboard.py
+++ b/xrviz/dashboard.py
@@ -40,7 +40,7 @@ class Dashboard(SigSlot):
         self._register(self.plot_button, 'plot_clicked', 'clicks')
         self.connect('plot_clicked', self.create_plot)
 
-        self._register(self.control.coord_setter.set_coord_button, 'set_coords', 'clicks')
+        self._register(self.control.coord_setter.coord_selector, 'set_coords')
         self.connect("set_coords", self.set_coords)
 
         self.control.displayer.connect('variable_selected', self.check_is_plottable)
@@ -214,7 +214,7 @@ class Dashboard(SigSlot):
         # in coord_selector.value
         self.data = self.data.reset_coords()
         indexed_coords = set(self.data.dims).intersection(set(self.data.coords))
-        new_coords = set(self.control.coord_setter.coord_selector.value).union(indexed_coords)
+        new_coords = set(args[0]).union(indexed_coords)
         self.data = self.data.set_coords(new_coords)  # this `set_coords` belongs to xr.dataset
         self.control.set_coords(self.data)
 
--- a/xrviz/dashboard.py
+++ b/xrviz/dashboard.py
@@ -45,8 +45,6 @@ class Dashboard(SigSlot):
 
         self.control.displayer.connect('variable_selected', self.check_is_plottable)
 
-        self.control.style.connect('colormap_limits', self.change_limits)
-
         self.panel = pn.Column(self.control.panel,
                                self.plot_button,
                                self.output)
@@ -76,7 +74,18 @@ class Dashboard(SigSlot):
                           'colorbar': self.kwargs['colorbar']}
             colormap_limits = self.kwargs['colormap_limits']
             color_scale = self.kwargs['color_scale']
-            sel_data = self.select_data()
+            dims_to_agg = self.kwargs['dims_to_agg']
+            sel_data = self.data[self.var]
+
+            for dim in dims_to_agg:
+                if self.kwargs[dim] == 'count':
+                    sel_data = (~ sel_data.isnull()).sum(dim)
+                else:
+                    agg = self.kwargs[dim]
+                    sel_data = getattr(sel_data, agg)(dim)
+
+            if self.var in list(sel_data.coords):  # When a var(coord) is plotted wrt itself
+                sel_data = sel_data.to_dataset(name=f'{sel_data.name}_')
 
             color_range = {sel_data.name: (sel_data.quantile(colormap_limits[0]),
                                                sel_data.quantile(colormap_limits[1]))}
@@ -121,6 +130,7 @@ class Dashboard(SigSlot):
             selection[dim] = self.index_selectors[i].value
         x = self.kwargs['x']
         y = self.kwargs['y']
+        dims_to_agg = self.kwargs['dims_to_agg']
         graph_opts = {'x': x,
                       'y': y,
                       'title': self.var,
@@ -130,24 +140,10 @@ class Dashboard(SigSlot):
                       'colorbar': self.kwargs['colorbar']}
         colormap_limits = self.kwargs['colormap_limits']
         color_scale = self.kwargs['color_scale']
-        sel_data = self.select_data()
 
-        color_range = {sel_data.name: (sel_data.quantile(colormap_limits[0]),
-                                       sel_data.quantile(colormap_limits[1]))}
+        sel_data = self.data[self.var]
 
-        sel_data = sel_data.sel(**selection, drop=True)
-        if color_scale is not 'linear':
-            sel_data = getattr(numpy, color_scale)(sel_data)  # Color Scaling
-        assign_opts = {dim: self.data[dim] for dim in sel_data.dims}
-        graph = sel_data.assign_coords(**assign_opts).hvplot.quadmesh(**graph_opts).redim.range(**color_range)
-        self.output[0] = graph
-
-    def select_data(self):
-        self.kwargs = self.control.kwargs
-        var = self.kwargs['Variables']
-        sel_data = self.data[var]
-
-        for dim in self.kwargs['dims_to_agg']:
+        for dim in dims_to_agg:
             if self.kwargs[dim] == 'count':
                 sel_data = (~ sel_data.isnull()).sum(dim)
             else:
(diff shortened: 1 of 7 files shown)


xrviz/dashboard.py
```diff
@@ -81,8 +88,13 @@ def create_plot(self, *args):
             if self.var in list(sel_data.coords):  # When a var(coord) is plotted wrt itself
                 sel_data = sel_data.to_dataset(name=f'{sel_data.name}_')
 
+            color_range = {sel_data.name: (sel_data.quantile(colormap_limits[0]),
+                                               sel_data.quantile(colormap_limits[1]))}
+            if color_scale is not 'None':
+                sel_data = getattr(numpy, color_scale)(sel_data)  # Color Scaling
```
martindurant: If we include `1 / x` as a scaling, this will not work. You could have made a simpler dict
```
{
 "linear": lambda x: x,
 "mean": np.mean,
 ...
  "1/x": lambda x: 1 / x
}[color_scale]
```

hdsingh: `1/x` is already present as a scaling method using [`numpy.reciprocal`](https://docs.scipy.org/doc/numpy/reference/generated/numpy.reciprocal.html) and is available as `reciprocal` in color_scale options.

martindurant: My mistake :)
It's not the name I would have guessed for it.

Note that you are calling numpy functions here, but actually they get converted to xarray functions ([example](http://xarray.pydata.org/en/stable/generated/xarray.DataArray.mean.html)). I don't know where reciprocal is implemented in xarray.

hdsingh:
In order to find lower and upper color limits, before creating a plot, we need to find the data selected according to the selections made in the fields panel. This `selected_data` would be updated upon:
- selection of new variable 
  or
- upon change of value of any aggregation option 
 or 
- setting/restting coords 

This needs to be done because the data selected/computed upon aggregation is different so its max and min values (color limits) will also change. 
Another point to note: color_limits do not change on just changing x and y, they change only upon change in aggregations options.
It is also better to find the `selected_data` once and use it again rather than finding it everytime upon changing values in colormap_limits.

As of now `cmap lower limit` and `cmap upper limit` are not linked with slider but are displaying the correct max and min values  on:
- selection of a new variable
- any change in aggregation.
- Setting/resetting coords

For now the updates are also being displayed in terminal.
I will soon do the final step of linking `cmap lower/upper limit` with slider.

rsignell-usgs:
We definitely don't want to read all the data to set the color limits, as some datasets may very large (e.g. I have a dataset with 200,000 time steps)

martindurant:
So would it make sense to go back to @rsignell-usgs 's original idea, and have two text boxes which only get passed as parameters if they contain a valid number? The boxes should then be cleared each time one of the changes above happens. In the case that a box is blank and we have selectors, at plot time we will want to get the max and/or min, to ensure consistent colouring across all selections.

hdsingh:
@rsignell-usgs @martindurant This is what I am going to implement. Please clarify if I am right.

We will only have two text boxes and no slider. These two boxes would initially be empty. Only when we create plot for a variable (after whatever selections the user has made), these would be filled from the same data that is being passed to plot creating functions. After plot creation, user can adjust text boxes values anywhere between the actual max and min and the colormap would be adjusted according to these values on creating plot again.

rsignell-usgs:
We should get the min/max from whatever plot has been selected, or the 1st time step, and just use that for the animation.

martindurant:
> or the 1st time step

I can imagine that going very wrong

hdsingh:
This is how style panel appears when no variable has been plotted.
Screenshot from 2019-06-30 14-07-12

Now the cmap change happens only if:
1. New variable is selected
2. Aggregations are changed

Now the calculation of color limits happens only once upon clicking PLOT on the entire selection using `sel_data.quantile`. These calculated values are used later on every time unless user has provided new cmap limits.

I think we should not have two different options because:
1. If the first time step has very different(less, more or null) values on average, the rest of selections/time steps will not be color balanced.
2. We are calculating limits only once and it does not takes virtually any time as compared to other aggregation functions, so is as much fast.
3. Having two separate  options will complicate style panel.

One more interesting thing I found:
It might be better if we don't apply validation on lower and upper cmap limits and leave it up to the user to decide because: 
1. There are no errors.
2. Some intereseting colormap (Color inversion, which could act as a feature)

Simple:
Screenshot from 2019-06-30 14-10-25

Inverted:
Screenshot from 2019-06-30 14-10-14

xrviz/dashboard.py
```diff
@@ -81,8 +98,25 @@ def create_plot(self, *args):
             if self.var in list(sel_data.coords):  # When a var(coord) is plotted wrt itself
                 sel_data = sel_data.to_dataset(name=f'{sel_data.name}_')
 
+            c_low_val = self.kwargs['cmap lower limit']
+            c_upp_val = self.kwargs['cmap upper limit']
+
+            # It is better to set initial values as 0.1,0.9 rather than 0,1 i.e. (min, max)
```
martindurant: This is, of course, a matter of opinion. Quantiles probably take considerably longer to calculate.

xrviz/dashboard.py
```diff
@@ -81,8 +98,25 @@ def create_plot(self, *args):
...
+            # to get a color balanced graph
+            c_lim_lower = float(c_low_val) if c_low_val else sel_data.quantile(0.1)
```
martindurant: I expect `quantile` can take a list argument, to compute both values in one step

xrviz/dashboard.py
```diff
@@ -81,8 +98,25 @@ def create_plot(self, *args):
...
+            c_lim_upper = float(c_upp_val) if c_upp_val else sel_data.quantile(0.9)
+
+            color_range = {sel_data.name: (c_lim_lower, c_lim_upper)}
+
+            if not c_low_val:  # if initial values are empty
```
martindurant: 0? You called `float` above, which will error on black or incorrect input.

xrviz/dashboard.py
```diff
@@ -81,8 +98,25 @@ def create_plot(self, *args):
...
+                self.control.style.lower_limit.value = str(c_lim_lower.values.round(5))
+                self.control.style.upper_limit.value = str(c_lim_upper.values.round(5))
+
+            if color_scale is not 'linear':
+                sel_data = getattr(numpy, color_scale)(sel_data)  # Color Scaling
```
martindurant: So is this in the original or scaled values?

xrviz/dashboard.py
```diff
@@ -81,8 +98,25 @@ def create_plot(self, *args):
...
+
             assign_opts = {dim: self.data[dim] for dim in sel_data.dims}
-            graph = sel_data.assign_coords(**assign_opts).hvplot.quadmesh(**graph_opts)
+            graph = sel_data.assign_coords(**assign_opts).hvplot.quadmesh(**graph_opts).redim.range(**color_range)
```
martindurant: I am surprised it is not a `clim=` argument to `quadmesh`. Perhaps `.redim` exists to be able to change the colours without recalculating the grid (but we don't make use of the intermediate object)

xrviz/dashboard.py
```diff
@@ -139,9 +178,25 @@ def create_indexed_graph(self, *args):
         if sel_data.name in self.data.coords:
                 sel_data = sel_data.to_dataset(name=f'{sel_data.name}_')
 
+        c_low_val = self.kwargs['cmap lower limit']
```
martindurant: I note that both the kwargs block (L160, above) and this colour-limit stuff is repeated.

xrviz/style.py
```diff
@@ -0,0 +1,38 @@
...
+        self.height = pn.widgets.IntSlider(name='height', value=300, start=100, end=1200)
+        self.width = pn.widgets.IntSlider(name='width', value=700, start=100, end=1200)
+        self.cmap = pn.widgets.Select(name='cmap', value='Inferno',
+                                      options=list_cmaps(provider='bokeh', reverse = False))
+        self.colorbar = pn.widgets.Checkbox(name='colorbar', value=True)
+        # colormap_limits
+        self.lower_limit = pn.widgets.TextInput(name='cmap lower limit', width=140)
+        self.upper_limit = pn.widgets.TextInput(name='cmap upper limit', width=140)
+
+        scaling_ops = ['linear', 'cos', 'exp', 'log', 'reciprocal',
```
martindurant: please remove trig functions

xrviz/dashboard.py
hdsingh: This is what I mean here by color balanced plot. I have tried setting `cmap lower limit` and `cmap upper limit` to  `sel_data.min()` and `max` and the graphs that we get are color imbalanced as follows:

Screenshot from 2019-07-02 07-34-06

Screenshot from 2019-07-02 07-35-46



If min or max is an outlier, it will result in the entire graph having almost same color.

xrviz/dashboard.py
hdsingh: As discussed earlier [here in Set/reset Coords](https://github.com/intake/xrviz/pull/7#discussion_r294266150), var_coords and var_dims are handled separately. For the time being, we should have it like this only(simpler) since projection panel is also using same code. Merging the code of  two cases here will result in more conflict in Projection Panel. As of now it is better to have these separate to understand two cases easily. Code cleanup would be done after all sections are complete.

xrviz/dashboard.py
hdsingh: I have handled this case. Thanks!

xrviz/dashboard.py
hdsingh: Now both values are being computed in one step.

xrviz/dashboard.py
hdsingh: Color_scaling in done on the original data. `Cmap Limits` are calculated using original and not scaled data.

hdsingh:
I have added Option 2: check box `compute min/max from all data`. I have also implemented change in cmap limits according to color scale. Since the scaled data values are different from orignal values, it might be better to use scaled values in cmap limits. For ex:

Without scaled values in cmap limits:
Screenshot from 2019-07-02 13-45-57

With scaled values in cmap limits:
Screenshot from 2019-07-02 13-46-55

xrviz/dashboard.py
hdsingh: This has been changed now.

xrviz/dashboard.py
martindurant: True in this case, but dependent on the data. We *could* allow selections to auto-fill the colour values at min/max, 10/90, 25/75... but I feel we are running out of space.

xrviz/dashboard.py
martindurant: I didn't follow your reasoning here, perhaps this is a good point for discussion.

hdsingh: Initially on color_scaling, quantiles(0.1,0.9) of actual data was filled in `cmap limits`. It resulted in completely dark graph for most of the color scales, since on scaling the data values change. Now the values of `cmap limits` will become empty on selection of new scale. After clicking plot, the values are filled according to quantiles of scaled values due to which graph will have proper colors. (as explained [below](https://github.com/intake/xrviz/pull/13#issuecomment-507574327)).

martindurant:
@rsignell-usgs , can I ask you to review this?

xrviz/dashboard.py
```diff
@@ -67,8 +80,13 @@ def create_plot(self, *args):
...
                           'y': y,
                           'title': self.var,
-                          'cmap': 'Inferno'}
+                          'height': self.kwargs['height'],
+                          'width': self.kwargs['width'],
+                          'cmap': self.kwargs['cmap'],
+                          'colorbar': self.kwargs['colorbar']}
+            color_scale = self.kwargs['color_scale']
             dims_to_agg = self.kwargs['dims_to_agg']
+            use_all_data = self.kwargs['compute min/max from all data']
```
martindurant: To be revisited later: I think right now this is actually 10/90%ile of the whole dataset if True, or min/max of first slice. I expect users to have opinions about what options should be presented here.

xrviz/dashboard.py
```diff
@@ -81,8 +99,31 @@ def create_plot(self, *args):
...
+                sel_data = getattr(numpy, color_scale)(sel_data)  # Color Scaling
+
+            if not use_all_data:
+                # sel the values at first step, to use for cmap limits
+                sels = {dim: 0 for dim in self.kwargs['remaining_dims']}
+                sel_data_for_cmap = sel_data.isel(**sels, drop=True)
+            else:
+                sel_data_for_cmap = sel_data
+
+            c_low_val, c_upp_val = self.kwargs['cmap lower limit'], self.kwargs['cmap upper limit']
```
martindurant: Could we call them cmax and cmin?

xrviz/dashboard.py
martindurant: ping; it would be good to have a comment on the usage here, which seems little odd to me

xrviz/dashboard.py
```diff
@@ -123,7 +164,13 @@ def create_indexed_graph(self, *args):
         graph_opts = {'x': x,
                       'y': y,
                       'title': self.var,
-                      'cmap': 'Inferno'}
+                      'height': self.kwargs['height'],
+                      'width': self.kwargs['width'],
+                      'cmap': self.kwargs['cmap'],
+                      'colorbar': self.kwargs['colorbar']}
+        color_scale = self.kwargs['color_scale']
+        use_all_data = self.kwargs['compute min/max from all data']
```
martindurant: To be revisited later: I think right now this is actually 10/90%ile of the whole dataset if True, or min/max of first slice. I expect users to have opinions about what options should be presented here.

xrviz/style.py
```diff
@@ -0,0 +1,42 @@
...
+        super().__init__()
+        self.height = pn.widgets.IntSlider(name='height', value=300, start=100, end=1200)
+        self.width = pn.widgets.IntSlider(name='width', value=700, start=100, end=1200)
+        self.cmap = pn.widgets.Select(name='cmap', value='Inferno',
+                                      options=list_cmaps(provider='bokeh', reverse = False))
+        self.colorbar = pn.widgets.Checkbox(name='colorbar', value=True)
+        # colormap_limits
+        self.lower_limit = pn.widgets.TextInput(name='cmap lower limit', width=140)
+        self.upper_limit = pn.widgets.TextInput(name='cmap upper limit', width=140)
+        self.use_all_data = pn.widgets.Checkbox(name='compute min/max from all data', value=True)
```
martindurant: To be revisited later: I think right now this is actually 10/90%ile of the whole dataset if True, or min/max of first slice. I expect users to have opinions about what options should be presented here.

xrviz/style.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.height = pn.widgets.IntSlider(name='height', value=300, start=100, end=1200)
+        self.width = pn.widgets.IntSlider(name='width', value=700, start=100, end=1200)
+        self.cmap = pn.widgets.Select(name='cmap', value='Inferno',
+                                      options=list_cmaps(provider='bokeh', reverse = False))
+        self.colorbar = pn.widgets.Checkbox(name='colorbar', value=True)
+        # colormap_limits
+        self.lower_limit = pn.widgets.TextInput(name='cmap lower limit', width=140)
```
martindurant: Is an absolute width necessary here? I imagine it will turn out differently for someone with a resolution very different from yours

martindurant:
@hdsingh , just a couple of questions - sorry they came out as individual comments rather than a review.

xrviz/style.py
hdsingh: I think it can be useful, since all we need is a box small enough to enter 1-2 digits usually and max 6-7 is values are large or minor decimal changes are needed. So it is better to have a box of width fixed for that.

xrviz/style.py
hdsingh: Yes, we will change this according to common opinion of users.

xrviz/style.py
martindurant: Yes, but the point is the actual visual size will depend on the user's system. It is better to use auto-sizing whenever possible. However, I will l leave it for now.

PR #14, opened 2019-06-26
Projection panel

hdsingh:
Screenshot from 2019-06-26 12-24-55

@rsignell-usgs @martindurant The following options are available in projection panel:
- basemap
- alpha(opacity of the graph wrt basemap)
- projection
- features

All options these are available only when user has checked `is_geo`. The geographical map would be generated only when both `x` and `y` are_coords. However for some of combinations such as (`sigma`,`time`) for variable `temp`, geographic map is not possible. So  it has been left upon to user to decide and select appropriately.


martindurant:
You should merge from master, since the names and order of the tabs changed in #12

xrviz/dashboard.py
```diff
@@ -1,7 +1,9 @@
 import panel as pn
 import xarray as xr
 import hvplot.xarray
-from cartopy import crs as ccrs
+import holoviews as hv
+from cartopy import crs
```
martindurant: Maybe the import should be deferred until it is needed

xrviz/utils.py
```diff
@@ -62,3 +62,12 @@ def callback(*events):
     source.param.watch(callback, ['value'], onlychanged=False)
     target = pn.Column(pn.Row(source.name, mark), source)
     return target
+
+
+projections_list = ['RotatedPole', 'Mercator', 'LambertCylindrical',
```
martindurant: Can this list be imported from somewhere? It would be a pain to have to keep it synced with the capabilities of cartopy. I would use alphabetic ordering here (`sorted()`).

hdsingh: I have `sorted` the list. I have checked but didn't found any source to import projections list.

xrviz/dashboard.py
hdsingh: Corrected. Thanks!

rsignell-usgs:
@hdsingh I tried testing and got this:
```
(xrviz) C:\Users\rsignell\Documents\GitHub\xrviz\examples>python control_section.py
Traceback (most recent call last):
  File "control_section.py", line 3, in <module>
    from xrviz.dashboard import Dashboard
  File "c:\users\rsignell\documents\github\xrviz\xrviz\dashboard.py", line 6, in <module>
    import geoviews.feature as gf
ModuleNotFoundError: No module named 'geoviews'
```

hdsingh:
@rsignell-usgs I will add `geoviews` in the requirements.txt.

rsignell-usgs:
Great!  Looks like you still are having problems with projections other than `PlateCarree`.   Did you get a chance to look at the geoviews documentation to know what projects should work?   As @martindurant suggested, you could try some simple tests outside of this app to see what works (and raise an issue on geoviews if something that is supposed to work isn't working!)

hdsingh:
As of now I have not looked into the reason behind improper functioning of projections. I will soon test the projections in a separate notebook and share my findings with you.

xrviz/utils.py
martindurant: Unfortunately, it requires importing `crs`, which would be better avoided, if possible. I'm not sure if we've decided to depend upon cartopy/geoviews or not. Does hvplot require them?
```
[p for p in dir(crs) if inspect.isclass(getattr(crs, p)) and issubclass(getattr(crs, p), crs.Projection) and not p.startswith('_')]
```

martindurant:
I am not certainly if we want geoviews as a required dependency. Importing only when required, and giving a sensible error message when needed but not found could be nicer.

hdsingh:
[hvplot/environment.yml](https://github.com/pyviz/hvplot/blob/master/environment.yml) does not lists cartopy and geoviews as a dependency. So we can import these only if the user sets `is_geo` and plots,  an error message would be raised to install if not found.

martindurant:
If cartopy is missing, should the projections panel be omitted?

hdsingh:
> If cartopy is missing, should the projections panel be omitted?

@rsignell-usgs @martindurant So what is final regarding this? Shall I proceed further with this panel?

martindurant:
> Shall I proceed further with this panel?

Certainly. I am saying that you should check at runtime if cartopy is importable and only create this tab if it is.

xrviz/utils.py
hdsingh: Corrected. Thanks!

hdsingh:
@rsignell-usgs @martindurant Now we are able to project data. The default `crs` is `None` which is able to correctly create plot. Please have a look.

hdsingh:
This is how the projection panel appears now. For the selection:
`projection`: Orthographic, 
`central_latitude`: -78,
`central_longitude`: 43

Screenshot from 2019-07-02 17-33-13

Upon zoom out, we can see correct projection with  these parameters.

martindurant:
Does this mean that you are having to find the required extra parameters, and create input widgets, for each of the projection types?

hdsingh:
There are several parameter available for each of the projection out of which I am considering only `central_latitude` and `central_longitude` (since they are most important). Yes, for each of projection I am finding if these parameters exist and then create a widget for them dynamically. Creation of new widget happens upon change in value of  `projection`.

hdsingh:
I discussed plotting basemap along with projections in [pyviz - Gitter](https://gitter.im/pyviz/pyviz)
>Just to confirm: Is it possible to display tile sources according to projections in geoviews/hvplot? Ex:
>```python
>air_ds = xr.tutorial.open_dataset('air_temperature').load()
>
>graph = air_ds.hvplot.quadmesh('lon', 'lat', 'air', projection=ccrs.Orthographic(-90, 30),
>                                global_extent=True, width=600, height=540, cmap='viridis') * >gv.feature.coastline
>graph * gv.tile_sources.CartoDark
>```
>This displays entire tile in background, but can it be only on and according to projected surface?

>James A. Bednar  Jul 01 23:21
I believe that Bokeh tile sources can only be displayed in GOOGLE_MERCATOR projections, which is a limitation that would be inherited by geoviews and hvplot when those use a Bokeh backend.

 and found that  that only ccrs.Mercator and ccrs.GOOGLE_MERCATOR work with gv.tile_sources. So only two projections would not be of much use to us.

I think in order to include basemaps, at one time user can have only one of basemap or (projection, crs) enabled.
So a check box which turn would turn on basemap and turn off projection,crs could be a solution for having basemaps.

hdsingh:
Now we can choose `basemap` (when `show_map` is on) or choose `projection`,`crs` when `show_map` is off.

Screenshot from 2019-07-03 08-21-41

hdsingh:
I have included metpy. Now `x` and `y` are displaying correct defaults.

xrviz/control.py
```diff
@@ -29,14 +29,21 @@ class Control(SigSlot):
     def __init__(self, data):
         super().__init__()
         self.data = data
+        self.has_cartopy = self.is_cartopy_installed()
```
martindurant: This should probably be global: not a class method not an instance attribute

requirements.txt
```diff
@@ -3,3 +3,5 @@ hvplot
 xarray
 netcdf4
 cartopy
+geoviews
+metpy
```
martindurant: I imagine this has heavy requirements. Can it be treated like cartopy, only used if it's available?

xrviz/dashboard.py
```diff
@@ -31,6 +31,7 @@ def __init__(self, data):
         if not isinstance(data, xr.core.dataarray.DataWithCoords):
             raise ValueError("Input should be an xarray data object, not %s" % type(data))
         self.set_data(data)
+        self.cartopy_installed()
```
martindurant: Yeah, so this makes duplication: would be better to have the import logic in utils module, and import, rather than repeatedly calling methods.

xrviz/dashboard.py
```diff
@@ -82,7 +111,12 @@ def create_plot(self, *args):
                 sel_data = sel_data.to_dataset(name=f'{sel_data.name}_')
 
             assign_opts = {dim: self.data[dim] for dim in sel_data.dims}
-            graph = sel_data.assign_coords(**assign_opts).hvplot.quadmesh(**graph_opts)
+            graph = sel_data.assign_coords(**assign_opts).hvplot.quadmesh(**graph_opts).opts(active_tools=['wheel_zoom', 'pan'])
+
+            if self.has_cartopy and is_geo:
+                graph = feature_map * graph
```
martindurant: There may not be any features selected

xrviz/dashboard.py
```diff
@@ -140,7 +175,7 @@ def create_indexed_graph(self, *args):
 
         sel_data = sel_data.sel(**selection, drop=True)
         assign_opts = {dim: self.data[dim] for dim in sel_data.dims}
-        graph = sel_data.assign_coords(**assign_opts).hvplot.quadmesh(**graph_opts)
+        graph = sel_data.assign_coords(**assign_opts).hvplot.quadmesh(**graph_opts).opts(active_tools=['wheel_zoom', 'pan'])
```
martindurant: Box zoom

xrviz/dashboard.py
```diff
@@ -122,7 +156,8 @@ def create_indexed_graph(self, *args):
         dims_to_agg = self.kwargs['dims_to_agg']
         graph_opts = {'x': x,
                       'y': y,
-                      'title': self.var}
+                      'title': self.var,
+                      'cmap': 'Inferno'}
```
martindurant: Why the absolute cmap value here?

xrviz/dashboard.py
```diff
@@ -202,3 +230,14 @@ def check_is_plottable(self, var):
...
+
+    def cartopy_installed(self):
+        try:
+            from cartopy import crs as ccrs
+            import geoviews as gv
+            import geoviews.feature as gf
+            global ccrs, gv, gf
+            self.has_cartopy = True
+        except:
+            print("Install Cartopy to view Projection Panel.")
```
martindurant: This would be a warning or a logging command, never a print. Only use print for short-term debugging.

xrviz/fields.py
```diff
@@ -85,19 +87,27 @@ def change_y(self, value=None):
...
+        self.are_var_coords = self.check_are_var_coords()
         self.change_dim_selectors()
 
     def change_dim_selectors(self, *args):
+        self.are_var_coords = self.check_are_var_coords()
         self.agg_selectors.clear()
-        used_opts = [self.x.value, self.y.value]
+        x = self.x.value
+        y = self.y.value
+        used_opts = set([x, y])
```
martindurant: `{x. y}`

xrviz/fields.py
```diff
@@ -134,3 +145,20 @@ def set_coords(self, data, var):
...
+        x = self.x.value
+        y = self.y.value
+        return True if x in var_coords and y in var_coords else False
+
+    def guess_x_y(self, var):
+        try:
+            parsed_var = self.data.metpy.parse_cf(var)
+            x, y = parsed_var.metpy.coordinates('x', 'y')
+            return [coord.name for coord in (x, y)]
+        except:
```
martindurant: When does this fail?

xrviz/projection.py
```diff
@@ -0,0 +1,113 @@
...
+                                                 value=False)
+        self.proj_params = pn.Row()
+
+        feature_ops = ['None', 'borders', 'coastline', 'grid', 'land', 'lakes',
+                       'ocean', 'rivers']
+        self.features = pn.widgets.MultiSelect(name='features',
+                                               options=feature_ops,
+                                               value=feature_ops[1:])
+
+        self._register(self.is_geo, 'is_geo_value')
```
martindurant: Prefer the names to be verbs, since they represent actions, so "geo_changed" in this case, "geo_disabled" for the next.

xrviz/tests/test_control.py
```diff
@@ -10,4 +10,4 @@ def control(data):
 
 def test_control_initial(control):
     tabs = [tab.name for tab in control.tabs]
-    assert tabs == ['Fields', 'Set Coords']
+    assert tabs == ['Fields', 'Projection', 'Set Coords']
```
martindurant: This is only the case under your system and Travis, when we've controlled the environment. For other people it would fail, if they don't have cartopy.

xrviz/utils.py
```diff
@@ -62,3 +62,12 @@ def callback(*events):
     source.param.watch(callback, ['value'], onlychanged=False)
     target = pn.Column(pn.Row(source.name, mark), source)
     return target
+
+
+def is_float(a):
+    # int, 'Nan' are also float
```
martindurant: also `inf` and `-inf`

xrviz/dashboard.py
hdsingh: `cmap` option is present in style panel. It would be replaced once style panel is merged (while resolving conflicts).

xrviz/dashboard.py
hdsingh: Even if we provide both `pan` and `box_zoom` only one of these (last in the list) would be active. I have replaced `pan` with `box_zoom`.

xrviz/fields.py
hdsingh: This fails when:
1. `lat`,`lon` are not available as coordinates.
2. For example dataset: when `mask` is also set as coordinate.

I have added a comment.

xrviz/dashboard.py
hdsingh: In that case `feature_map` would be empty i.e. `gv.Overlay()`. There would be no error.

requirements.txt
hdsingh: I have handled this.

xrviz/dashboard.py
hdsingh: I have tried but not able to figure out a way to import `cartopy`, `geoviews` in utils and make these availble globally in dashboard. So `cartopy_geoviews_installed` is defined in dashboard and `has_cartopy` is also now global in dashboard.

xrviz/dashboard.py
hdsingh: Corrected. Now using `warnings.warn`.

xrviz/tests/test_control.py
hdsingh: I has been handled now.

hdsingh:
@rsignell-usgs @martindurant I have resolved the conflicts and have made changes according to the above review. Please have a look.

xrviz/dashboard.py
martindurant: We *may* get annoyed feedback on this, because users without cartopy will see this every time. Might need to change later.

hdsingh:
@martindurant I think this could be merged now. I will create a separate PR for test related to style and projections panel. Please let me know if anything else needs to be changed here.

examples/display_section.py
```diff
@@ -1,8 +1,9 @@
 import xarray as xr
 import xrviz
 from xrviz.display import Display
+from xrviz.sample_data import great_lakes
```
martindurant: Same as above.
In fact, I'm not sure you need this module at all any more.
I would prefer that you implement an example module within the xrviz package, such that
```
import xrviz; xrviz.example()
```
runs the full interface, but that the example data is *not* loaded unless either this is called, or it is ecxplicitly imported.

xrviz/control.py
```diff
@@ -30,6 +31,7 @@ class Control(SigSlot):
     def __init__(self, data):
         super().__init__()
         self.data = data
+        self.has_cartopy = cartopy_geoviews_installed()
```
martindurant: Why the need for the attribute? Can't you refer to the global variable everywhere?

xrviz/dashboard.py
```diff
@@ -31,6 +33,7 @@ def __init__(self, data):
         if not isinstance(data, xr.core.dataarray.DataWithCoords):
             raise ValueError("Input should be an xarray data object, not %s" % type(data))
         self.set_data(data)
+        cartopy_geoviews_installed()
```
martindurant: It is odd to have this run every time.
Better, I think, would be to have a compatibility module which we import all the optional modules from, so that the checks happen only once
```
from .compatability import ccrs, gv, g, has_cartopy
```

and in the module:
```
try:
    from cartopy import crs as ccrs
    import geoviews as gv
    import geoviews.feature as gf
    has_cartopy = True
except ImportError:
    logging.debug('...')
```

That way you only call it once, and you can reuse this in other places. The downside would be that this happens on import of xrviz *unless* we contrive a way to do the import and keep hold of the references during `__init__`. I find the simple method above is much easier to understand, so probably worth any small import time cost.

xrviz/dashboard.py
```diff
@@ -271,3 +302,20 @@ def check_is_plottable(self, var):
...
+    Projection panel would be displayed only if both present.
+    """
+    global has_cartopy
+    try:
+        from cartopy import crs as ccrs
+        import geoviews as gv
+        import geoviews.feature as gf
+        has_cartopy = True
+        global ccrs, gv, gf
+    except:
```
martindurant: `except ImportError` - because other exceptions should show up

xrviz/fields.py
```diff
@@ -136,3 +148,32 @@ def set_coords(self, data, var):
...
+            return [None, None]
+
+
+def has_metpy():
+    global mpcalc
+    try:
+        import metpy.calc as mpcalc
+    except:
+        warnings.warn("""Install Metpy using `pip install metpy` or 
+                      `conda install -c conda-forge metpy pint pooch --no-deps`
```
martindurant: Are you sure about this line?

xrviz/fields.py
```diff
@@ -136,3 +148,32 @@ def set_coords(self, data, var):
...
+                      to automatically guess values of x,y""",
```
martindurant: `automatically guess values of x,y` - there is no context, the user won't know why this appeared. Wording could be like
"
Metpy is not installed, so xrviz cannot automatically guess the best dimensions for output x and y axes.
To install metpy, you can execute the following:
...
"

xrviz/tests/test_control.py
```diff
@@ -10,4 +11,8 @@ def control(data):
 
 def test_control_initial(control):
     tabs = [tab.name for tab in control.tabs]
-    assert tabs == ['Set Coords', 'Variables', 'Axes', 'Style']
+    if cartopy_geoviews_installed():
```
martindurant: Only one of these will ever be triggered in a given test run.

xrviz/utils.py
```diff
@@ -65,9 +65,18 @@ def callback(*events):
...
-    # int, 'Nan' are also float
+    # int, 'Nan', 'inf' and '-inf' are also float
     try:
         float(a)
         return True
     except:
         return False
+
+
+def cartopy_geoviews_installed():
```
martindurant: It is somewhat dangerous to repeat code for the purposes of testing. If the test above breaks, it may not be obvious if the error is in the main code or here.

xrviz/dashboard.py
hdsingh: This method works perfect and solves earlier problems. Thanks!

xrviz/fields.py
hdsingh: Changed according to above comment. Thanks!

examples/control_section.py
```diff
@@ -1,8 +1,9 @@
 import xarray as xr
 import xrviz
 from xrviz.dashboard import Dashboard
+from xrviz.sample_data import great_lakes
```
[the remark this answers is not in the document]
hdsingh: Corrected according to 2nd method.

xrviz/utils.py
hdsingh: Changed. Now compatibility module handles these optional imports.

xrviz/fields.py
hdsingh: Changed. Thanks!

xrviz/tests/test_control.py
hdsingh: Since this is dependent on presence on optional dependencies, only one of these will be triggered.

examples/display_section.py
hdsingh: I have deleted `control_section` and `display_section` since they are no longer required. `import xrviz; xrviz.example()`  will be used.

hdsingh:
@martindurant I have made changes according to the review. Please let me know if anything else needs to be done here.

rsignell-usgs:
I'm back from vacation, and just unburied from email.  Can I test this a bit before we merge?

xrviz/tests/test_control.py
martindurant: Yes. My point here was, I'm not sure how useful this is as a test. More typical might be to skip if the module import fails, and only check the True branch.

rsignell-usgs:
oops, I should have added [skip ci].

rsignell-usgs:
I tried out the projection_panel PR here, and I'm impressed with the functionality.   I'm worried about new users and the additional complexity, but I don't have good ideas for improvement right now. 

A few comments:
- I understand the rationale for having `set_coords` be the first tab, but I really think `variables` should be the first tab, so that someone can just select a variable and click "plot" and get something showing in the GUI.
- I don't understand what the `crs` option does (maybe I just have to read back through the PR comments more)
- It's cool that the polar projections work.  Not sure why some of the land is colored blue and some yellow here though.  Is this a cartopy problem or our problem?

- I'm not sure what `metpy` is doing for us here.  I still got x=lat, y=lon by default.

rsignell-usgs:
Also, was just testing whether the user could switch between variables and have the settings stay pretty much the same, which would be nice, as this is a common use case. 

I tried selecting temperature, adding lon,lat as coordinates, checked `is_geo`, added the map, and then selected `air_u`.   Here's what I got, which is pretty strange:

hdsingh:
>I tried selecting temperature, adding `lon`,`lat` as coordinates, checked `is_geo`, added the map, and then selected air_u and clicked "plot".
> Here's what I got, which is pretty strange:

I followed the exact same steps and this didn't happened for me (I set only `lat` and `lon` as coord).
Here is the reason this happened: 
While setting coords, you must have added `mask`, due to which somehow Metpy is not able to figure out exact values of `x` and `y`. Since `lat`,`lon`, and `mask` have same dims, I think this is reasonable.

Metpy works for most of the cases, but when it fails, `x` and `y` are filled in alphabetical order.
Hence `x`,`y` have `lat`, `lon` which produces the exact same (strange) map.

hdsingh:
> *   I understand the rationale for having`set_coords`be the first tab, but I really think`variables`should be the first tab, so that someone can just select a variable and click "plot" and get something showing in the GUI.

Changed. Now `Variables` is first tab, `Set Coords` is second.

-----


> * I don't understand what the`crs`option does

As stated in [Geographic Data — hvPlot documentation](https://hvplot.pyviz.org/user_guide/Geographic_Data.html)
>The`crs=`argument specifies the_input_projection, i.e. it declares how to interpret the incoming data values. You can independently choose any_output_projection, i.e. how you want to map the data points onto the screen for display, using the`projection=`argument.

So  for this particular example data, `crs` doesn't have mush effect but will be useful for some others.

-----
>It's cool that the polar projections work. Not sure why some of the land is colored blue and some yellow here in this `NorthPolar` plot though. Is this a cartopy problem or our problem?

We are simply passing user selected arguments to hvplot, so must be something related to geoviews/cartopy.

hdsingh:
Since we have so much empty space in  Variable Panel, I have increased height of its sub sections.

rsignell-usgs:
We should try the HRRR dataset with the crs, to see if it projects correctly

hdsingh:
Now:
- `crs` and `projection` are independent.
- `projection` and `basemap` are mutually exclusive.

I have changed according to this. Now when `show_map` is on projection is disabled and vice-versa. `show_map` has no effect on `crs`.
I have tried the following code:
```python
import xarray as xr
from xrviz.dashboard import Dashboard

url = 'http://thredds-jetstream.unidata.ucar.edu/thredds/dodsC/grib/NCEP/HRRR/CONUS_2p5km/Best'
data = xr.open_dataset(url)

# To correctly project from metres to kms (since the data was initially in meters)
data['x'].values = data['x'].values*1000.
data['y'].values = data['y'].values*1000.

dashboard = Dashboard(data)
dashboard.show()
```

with these settings:
Variable: `Temperature_height_above_ground`
`is_geo`: True
`show_map`: True
`crs`: `LambertConformal`
It works fine. 

Screenshot from 2019-07-09 17-54-04


In order to correctly  project, when `crs` is selected, we need to have proj_params: `central_lon`, `central_lat` for `crs` as well.
Shall I add these parameters for `crs` also?

hdsingh:
> Shall I add these parameters for `crs` also?

After adding parameter for `crs` and using the interface I realised there is no need for `crs` since we can easily specify correct orientation using `projection`.

Ex: For HRRR datset
Screenshot from 2019-07-10 00-33-45

martindurant:
Am I wrong, or is that image misaligned with the map? The image seems to have been compressed relative to the map.

rsignell-usgs:
@hdsingh that HRRR model is definitely *not* in the correct location.   

To get this right we need the correct globe, as well as the right central_latitude, central_longitude, and standard_parallels, as in: 
                      
https://github.com/reproducible-notebooks/HRRR_Dashboard/blob/master/HRRR_Dashboard.ipynb

hdsingh:
So would it be okay to have widgets for central_latitude, central_longitude, and standard_parallels and globe as well for `crs`? (Besides central_latitude and central_longitude for `projection`)

martindurant:
I think I'm missing something, what does it mean to have the right globe?

hdsingh:
Globe Defines an ellipsoid and, optionally, how to relate it to the real world as stated in [cartopy documentation](https://scitools.org.uk/cartopy/docs/v0.15/crs/index.html#cartopy.crs.Globe). I am looking in detail how it effects.

rsignell-usgs:
Yeah, this is an entire field:  google "geodesy" !

But basically, there are some commonly used models of the earth.  A sphere with a certain radius is common for modelers, while more accurate models are ellipsoids like WGS84.

hdsingh:
This displays correct projection, even without globe. I think adding globe and other parameters would overpopulate projection panel.

Screenshot from 2019-07-10 14-22-57

bokeh_plot

rsignell-usgs:
@hdsingh, that projection is still not right.    You *do* need the globe!   We can discuss at the meeting.

rsignell-usgs:
@hdsingh, I think the outcome from our weekly meeting was to merge this after:
1) there is a text box added (replacing the projection parameter widgets) to allow the user to supply a dictionary of projection parameters
2) the `show-map` widget is removed, and `None` is added to the `basemap` list

martindurant:
> there is a text box added

Can I amend this to one text box for each of the objects created (crs, projection)? As suggested on gitter, this could be a [LiteralInput](https://panel.pyviz.org/reference/widgets/LiteralInput.html) boxes of dict type.

hdsingh:
I will complete these 2 tasks as soon as possible (most probably by tomorrow).

xrviz/example.py
```diff
@@ -0,0 +1,29 @@
...
+    ----------
+    show: bool (True)
+        Whether to directly execute the interface. If True, a new browser tab
+        will be opened, and the function will block until interrupted. If
+        False, the Dashboard instance is returned without being executed.
+    """
+    from .sample_data import great_lakes
+
+    great_lakes = great_lakes.set_coords(['lat', 'lon'])
+    initial_params = {'projection_ops': {'is_geo': True,
```
hdsingh: I have added a way to initialise projection panel values by passing values as `dict`. As of now we can only  initialise projections panel. Please let me know if this is correct approach before I add this method in other panels as well.

hdsingh:
Also I have added LiteralInput for `crs` and `projection` params, which is pre-filled with the arguments that can be passed to the selected projection.

xrviz/example.py
martindurant: I would prefer a name here which is identical to the tab title, "projection", in this case.
There is also an argument for having this flat, single-level, rather than nested, since no widget names are duplicated anywhere in the interface. Each panel can use only the keys which are relevant to its widgets.

rsignell-usgs:
@hdsingh , this new functionality is awesome!   I love that the lon/lat coordinates are automatically added, the "is_geo" box is prechecked, and the "none" in the basemap instead of the check box.      

The only thing I wish is that there was a way to get the default axes to be "x=lon", "y=lat" if we click "plot" on the geo page with "is_geo" enabled.   Is there any way to do that?

xrviz/projection.py
```diff
@@ -87,6 +87,15 @@ def setup(self, *args):
                 widget.disabled = disabled
         self.proj_params.disabled = disabled
 
+    def setup_initial_values(self, init_params):
+        is_geo_disabled = self.is_geo.disabled  # to restore state of is_geo
+        if init_params:
+            self.is_geo.disabled = False
```
martindurant: This pattern has come up a few times, why set and reset the value?
Simpler:
```
def setup_initial_values(self, init_params={}):
    for widget in [self.projection, self.proj_params]:   # or use strings here
        if widget.name in init_params:
            widget.value = init_params[widget.name]
```
I wasn't totally sure whether you wanted is_geo in this list or not. The simplest way not to automatically set it, is simply not to pass it in the init_params

martindurant:
> a way to get the default axes to be "x=lon", "y=lat"

I'm sure this method for prepopulating values will be implemented in all parts of the interface. To pick the right things in the case where no initial conditions are supplied is a little harder. Perhaps we should simply special-case for coordinates with names like "lon", "longitude", ... to come first

rsignell-usgs:
I think the metpy code that @hdsingh included has the logic necessary to assign the longitude coordinate to x and the latitude coordinate to y.   I think if this logic works on a particular dataset, we should populate these as the defaults if `is_geo` is enabled.

hdsingh:
>  Is there any way to do that?

Yes, I will now implement this setting user params method for all parts of the interface, according to simpler method.

hdsingh:
Now we can set initial values for all parts of the interface.

xrviz/projection.py
hdsingh: I have made modifications according to this method. Thanks!

hdsingh:
@rsignell-usgs @martindurant This PR looks complete to me now. Please have a look.

rsignell-usgs:
The labels say longitude on x axis, and latitude on y axis, but x=lat, y=lon are the default axes still:

hdsingh:
This is something related to `geoviews`. Reproducible example:
```python
import xarray
import hvplot.xarray
from geoviews import tile_sources as gvts
 
data = xarray.tutorial.open_dataset('air_temperature')

graph = data.air.hvplot.quadmesh(x='lat',y='lon') 
gvts.ESRI * graph
```
Screenshot from 2019-07-13 07-57-58
I am discussing about this in pyviz-Gitter.

rsignell-usgs:
@hdsingh that is interesting, but what I meant was that if we have the ability to identify *longitude* and *latitude*, via the metpy logic, we should make *longitude* the default for the `x` axis, and *latitude* the default for `y` axis.

hdsingh:
Can you please specify in detail what do you mean by "default"? 
Even if we don't pass (`x`,`y`) as (`lat`,`lon`) in `initial_params` in example, x and y would be lat, lon (because of `Set Coords` )

rsignell-usgs:
@hdsingh, now that I have actually looked at the latest example code, I see these are all being specified through the `init_params`.   I think I must have forgot to checkout the latest projection_panel branch after fetching.  Things are working fine now.    Let's change the `init_params` value for the example basemap to `OSM` and then I would vote for merging!

xrviz/compatibility.py
```diff
@@ -0,0 +1,25 @@
+import logging
```
martindurant: We should have our own logger
```
logger = logging.get_logger('xrviz')
```
and then, below
```
logger.debug(...)
```
so that users can filter only logging statements from this package.

xrviz/example.py
```diff
@@ -0,0 +1,39 @@
...
+
+    Parameters
+    ----------
+    show: bool (True)
+        Whether to directly execute the interface. If True, a new browser tab
+        will be opened, and the function will block until interrupted. If
+        False, the Dashboard instance is returned without being executed.
+    """
+    from .sample_data import great_lakes
+    initial_params = {'Variables': 'temp',  # Select Variable
```
martindurant: Yes, I think this is exactly how I imagined it. We will have to describe in the docs, in detail, the set of parameters that can show up here.

xrviz/fields.py
```diff
@@ -107,6 +118,15 @@ def change_dim_selectors(self, *args):
             self._register(agg_selector, agg_selector.name)
             self.agg_selectors.append(agg_selector)
 
+    def setup_initial_values(self, init_params={}):
+        for widget in [self.x, self.y]:
+            if widget.name in init_params:
+                widget.value = init_params[widget.name]
+
+        for dim_agg in self.agg_selectors:
```
martindurant: Combine with line above, `for widget in [self.x, self.y] + self.agg_selectors:` ?

xrviz/fields.py
```diff
@@ -136,3 +157,20 @@ def set_coords(self, data, var):
...
+        var_coords = list(self.data[self.var].coords)
+        x = self.x.value
+        y = self.y.value
+        return True if x in var_coords and y in var_coords else False
+
+    def guess_x_y(self, var):
+        try:
+            parsed_var = self.data.metpy.parse_cf(var)
+            x, y = parsed_var.metpy.coordinates('x', 'y')
+            return [coord.name for coord in (x, y)]
```
martindurant: Special case for longitude to x-axis goes here (can be next PR). Or should we ask metpy about clarification to get this right? Seems like their function should get the order right.

xrviz/templates/variable.html
```diff
@@ -112,8 +112,9 @@
 
         div.table_div {
           font-size: 13px;
-          overflow: scroll; 
-          height: 140px;
+          overflow: scroll;
```
martindurant: This style will surely change, don't worry too much. That you need to tinker with absolute point-measured like this i a bad indication that it won't display well in a range of resolutions. Same goes for the Panel widget sizes elsewhere.

xrviz/utils.py
```diff
@@ -65,9 +67,21 @@ def callback(*events):
...
 def is_float(a):
-    # int, 'Nan' are also float
+    # int, 'Nan', 'inf' and '-inf' are also float
     try:
         float(a)
         return True
     except:
         return False
+
+if has_cartopy:
```
martindurant: Take away the `if` - this function simply does not get called ever if cartopy is missing

xrviz/control.py
```diff
@@ -59,9 +75,14 @@ def set_coords(self, data):
         self.describer.set_coords(self.data, var)
         self.fields.set_coords(self.data, var)
 
+    def check_is_projectable(self, *args):
+        self.projection.is_geo.disabled = not self.fields.kwargs['are_var_coords']
```
martindurant: (should really be a method done within Projection - or otherwise a "signal" on Projection - generally I would dissuade against setting inner widgets' values directly)

rsignell-usgs:
The `metpy` `parse_cf` accesser has the capability to unambiguously determine the appropriate coordinates for the 'x' and 'y' directions (e.g. here `x=lon`, y=lat`).    We should use that if we can.

xrviz/fields.py
hdsingh: This was my initial approach but we `can only concatenate list (not "Column") to list`, since `self.agg_selectors` is a `pn.Column`.

hdsingh:
>  The `metpy` `parse_cf` accesser has the capability to unambiguously determine the appropriate...

We are using `metpy.parse_cf `. It works well for most of the cases. But for some of the cases in which other coordinates also have same dimensions as that of lat and lon, it is unable to determine. 
For example: In this great_lakes data, it is not able to identify correctly when `mask` is also present along with `lat`, `lon`.

xrviz/fields.py
hdsingh: I have asked this in [Metpy-Gitter](https://gitter.im/Unidata/MetPy#). Hopefully we will get a solution.

hdsingh:
I have also added `basemap` in `initial_params` in example.

rsignell-usgs:
@hdsingh this looks ready to merge to me. 
@martindurant what do you think?

rsignell-usgs:
@hdsingh I just noticed that the hover tool is not showing the correct lon/lat values, at least in my environment.    Do you have this same behaviour on your end?

hdsingh:
Yes I am also having same behaviour. I checked and found that this is also present in [GreatLakes_Dashboard](https://github.com/hdsingh/COAWST-ROMS_Dashboards/blob/master/GreatLakes_Dashboard.ipynb).

xrviz/fields.py
martindurant: `for widget in [self.x, self.y] + list(self.agg_selectors):`

hdsingh: Ohh! my fault. That was simple. Thanks!

hdsingh:
@martindurant Since this PR is complete, can you please merge this so that I can create new PRs for tests and series extraction?

martindurant:
OK, this is in.
Obviously, it leads to some further questions, but better move forward than stagnate here.

Issue #17, opened 2019-06-28
Colormap not found if installed from pip


milancurcic:
I just tried out the example in xrviz following your Medium blog post. Not a conda user, so installed from pip. If I try to plot a variable, I hit a ValueError due to missing colormap, see full traceback:

```
ERROR:tornado.application:Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x7f08a3475390>>, <Future finished exception=ValueError('Supplied cmap fire not found among matplotlib, bokeh, or colorcet colormaps.',)>)
Traceback (most recent call last):
  File "/home/milan/Work/xrviz/venv/lib/python3.6/site-packages/tornado/ioloop.py", line 743, in _run_callback
    ret = callback()
  File "/home/milan/Work/xrviz/venv/lib/python3.6/site-packages/tornado/ioloop.py", line 767, in _discard_future_result
    future.result()
  File "/home/milan/Work/xrviz/venv/lib/python3.6/site-packages/tornado/gen.py", line 748, in run
    yielded = self.gen.send(value)
  File "/home/milan/Work/xrviz/venv/lib/python3.6/site-packages/bokeh/server/session.py", line 70, in _needs_document_lock_wrapper
    result = yield yield_for_all_futures(func(self, *args, **kwargs))
  File "/home/milan/Work/xrviz/venv/lib/python3.6/site-packages/bokeh/server/session.py", line 191, in with_document_locked
    return func(*args, **kwargs)
  File "/home/milan/Work/xrviz/venv/lib/python3.6/site-packages/bokeh/document/document.py", line 1127, in wrapper
    return doc._with_self_as_curdoc(invoke)
  File "/home/milan/Work/xrviz/venv/lib/python3.6/site-packages/bokeh/document/document.py", line 1113, in _with_self_as_curdoc
    return f()
  File "/home/milan/Work/xrviz/venv/lib/python3.6/site-packages/bokeh/document/document.py", line 1126, in invoke
    return f(*args, **kwargs)
  File "/home/milan/Work/xrviz/venv/lib/python3.6/site-packages/bokeh/document/document.py", line 916, in remove_then_invoke
    return callback(*args, **kwargs)
  File "/home/milan/Work/xrviz/venv/lib/python3.6/site-packages/panel/viewable.py", line 653, in _change_event
    self.set_param(**self._process_property_change(events))
[...]

PR #18, opened 2019-06-28
Fix pip installation

hdsingh:
This includes installation instructions and changes for pip install. [Issue 17](https://github.com/intake/xrviz/issues/17) would be resolved with this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,23 @@
 # xrviz
 Interactive visualisation interface for Xarrays
+
+## Installation
+
+### Using Conda
+```
+conda install -c conda-forge xrviz
+pip install git+https://github.com/intake/xrviz --no-deps
+```
+
+### Using Pip
+```
+git clone https://github.com/intake/xrviz.git
+cd xrviz
+pip install .
+```
+
+### Usage
+```
+cd examples
+python control_section.py
+```
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ pip install git+https://github.com/intake/xrviz --no-deps
 
 ### Using Pip
 ```
-pip install pip+https://github.com/intake/xrviz.git .
+pip install git+https://github.com/intake/xrviz.git
 ```
 
 ### Usage
--- a/README.md
+++ b/README.md
@@ -11,9 +11,7 @@ pip install git+https://github.com/intake/xrviz --no-deps
 
 ### Using Pip
 ```
-git clone https://github.com/intake/xrviz.git
-cd xrviz
-pip install .
+pip install pip+https://github.com/intake/xrviz.git .
 ```
 
 ### Usage
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,3 @@ panel
 hvplot
 xarray
 netcdf4
-matplotlib
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,4 +2,4 @@ panel
 hvplot
 xarray
 netcdf4
-cartopy
+matplotlib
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -1,2 +1,23 @@
...
+
+### Using Conda
+```
+conda install -c conda-forge xrviz
+pip install git+https://github.com/intake/xrviz --no-deps
+```
+
+### Using Pip
+```
+git clone https://github.com/intake/xrviz.git
```
martindurant: these three lines are equivalent to `pip install pip+https://github.com/intake/xrviz.git`. Have you tested it in a fresh environment with the current requirements.txt?

README.md
```diff
@@ -1,2 +1,23 @@
...
+cd xrviz
+pip install .
+```
+
+### Usage
+```
+cd examples
+python control_section.py
```
martindurant: This should be available in the package, ideally to be invoked as
```
```

xrviz/dashboard.py
```diff
@@ -1,7 +1,6 @@
 import panel as pn
 import xarray as xr
 import hvplot.xarray
-from cartopy import crs as ccrs
```
hdsingh: `cartopy` is not being used anywhere in the master. 
`pip install cartopy` does not works. The only way to install it is using conda or building from source as stated in [Installing Cartopy — cartopy 0.17.0 documentation](https://scitools.org.uk/cartopy/docs/latest/installing.html) . In order to use projection panel, installation using conda would be the only way.

requirements.txt
```diff
@@ -2,4 +2,4 @@ panel
 hvplot
 xarray
 netcdf4
-cartopy
+matplotlib
```
martindurant: I know this came up in the context of colour-maps, but I find it very strange - we don't actually use matplotlib, and it's a *big* requirement. Surely holoviews doesn't need matplotlib, when it can equally well plot with bokeh or plotly.

xrviz/dashboard.py
martindurant: This is now imported elsewhere?

martindurant: OK, so the projection panel should be unavailable if cartopy is not installed, and we need to decide whether to include it in the conda requirements or not.

README.md
hdsingh: Yes I have tested installing via pip in a new environment, used the interface and ran tests(after installing pytest).

martindurant: We may wish to have one of the travis test runs which installs from pip, like Intake does: https://github.com/intake/intake/blob/master/.travis.yml#L28

xrviz/dashboard.py
hdsingh: Now `cartopy` is not being imported anywhere.

hdsingh:
The default cmap: `fire` was from `colorcet` or `matplotlib`.In style panel we will only use `bokeh` maps using 
```python
from holoviews.plotting import list_cmaps
cmaps = list_cmaps(provider='bokeh',  reverse = False)
```

martindurant:
> In style panel we will only use bokeh maps 

Sounds good. We probably had too many anyway. I suppose, we could list the matplotlib ones too *if* it is installed (but don't worry about it now).

README.md
hdsingh: I have tried to figure this out, structure of both travis files look a bit different to me. For this task I need your help. Could this be done in a separate PR?

hdsingh:
@martindurant Pip install will work now so can we merge this PR so that pip users can also have a look at the interface now?

martindurant:
OK, I'll follow up with a travis build thing

PR #20, opened 2019-06-30
xrviz.example()

hdsingh:
As mentioned [here](https://github.com/intake/xrviz/pull/18#discussion_r298709296), the command 
```
``` 
can now run example.

@martindurant Shall I also delete the examples folder or are there any plans to add more examples?

### DIFF
--- a/xrviz/__init__.py
+++ b/xrviz/__init__.py
@@ -1,5 +1,3 @@
-from . import sample_data
-
 from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
@@ -20,6 +18,7 @@ def example(show=True):
             ```
     """
     import xarray as xr
+    from . import sample_data
     from . import dashboard
     ds = sample_data.great_lakes
     dash = dashboard.Dashboard(ds)
--- a/xrviz/__init__.py
+++ b/xrviz/__init__.py
@@ -6,6 +6,19 @@ del get_versions
 
 
 def example(show=True):
+    """
+    Generates interface for the sample dataset.
+
+    Parameters
+    ----------
+    show: True (by default): Directly opens the interface in a browser
+          False : Returns the interface(dashboard) object
+            Usage:
+            ```
+            dash = xrviz.example(show = False)
+            dash.panel  # To display interface in Jupyter Notebook
+            ```
+    """
     import xarray as xr
     from . import dashboard
     ds = sample_data.great_lakes
--- a/xrviz/__init__.py
+++ b/xrviz/__init__.py
@@ -5,9 +5,9 @@ __version__ = get_versions()['version']
 del get_versions
 
 
-def example():
+def example(show=True):
     import xarray as xr
     from . import dashboard
     ds = sample_data.great_lakes
     dash = dashboard.Dashboard(ds)
-    dash.show()
+    return dash.show() if show else dash
--- a/xrviz/__init__.py
+++ b/xrviz/__init__.py
@@ -9,17 +9,16 @@ def example(show=True):
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


xrviz/__init__.py
```diff
@@ -3,3 +3,11 @@
 from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+
+def example():
+    import xarray as xr
+    from . import dashboard
+    ds = sample_data.great_lakes
```
martindurant: It would be great if `sample_data` were imported here for the first time, rather than on line 1, so that it doesn't get loaded in sessions that don't use it.

xrviz/__init__.py
```diff
@@ -3,3 +3,11 @@
 from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+
+def example():
```
martindurant: I suggest an optional parameter `show=True`, which, when False, causes `dash` to be returned rather than calling `.show()` on it. Then, you can include a test for this function, and of course docstrings.

xrviz/__init__.py
hdsingh: A user may want to do have a look at data, by doing `data = xrviz.smaple_data` from anywhere else so I think it is needed.

xrviz/tests/test_example.py
```diff
@@ -0,0 +1,7 @@
+from .. import example
+from xrviz.dashboard import Dashboard
+
+
+def test_dashboard_example():
+    dash = example(False)
```
hdsingh: I have only checked case where `show=False`. `show = True` directly opens in the browser and has type `None`.

xrviz/__init__.py
hdsingh: I have added these changes. Thanks!

xrviz/__init__.py
martindurant: I disagree, I think such a user should explicitly do `import xrviz.sample_data` (which, of course, we document)

xrviz/tests/test_example.py
martindurant: (`show=True` blocks and doesn't return at all. We could get fancy and do a times interrupt signal to test, but let's not).

xrviz/__init__.py
hdsingh: Corrected. Thanks!

xrviz/__init__.py
```diff
@@ -1,5 +1,25 @@
...
 del get_versions
+
+
+def example(show=True):
+    """
+    Generates interface for the sample dataset.
+
+    Parameters
+    ----------
+    show: True (by default): Directly opens the interface in a browser
```
martindurant: The indentation has gone wrong here. Should be like
```
    show: bool (True)
        Whether to directly execute the interface. If True, a new browser tab will be opened,
        and the function will block until interrupted. If False, the Dashboard instance is
        returned without being executed.
 ```

File: xrviz/__init__.py [hunk 5]
```diff
@@ -1,5 +1,25 @@
...
+            dash = xrviz.example(show = False)
+            dash.panel  # To display interface in Jupyter Notebook
+            ```
+    """
+    import xarray as xr
+    from . import sample_data
+    from . import dashboard
+    ds = sample_data.great_lakes
+    dash = dashboard.Dashboard(ds)
+    return dash.show() if show else dash
```
martindurant: Simpler would be
```
    if show:
        dash.show()
    return dash
```

hdsingh:
I think this could be merged now. Please let me know if there is something else that needs to be changed here.

martindurant:
Merging this now. I suggest a some point you update the README to show that this is now the recommended way to try the interface.

Issue #5, opened 2019-06-16
Install sample data following pyviz approach?


rsignell-usgs:
I tried to run an example and it failed to find the example data: 
```python-traceback
----> 3 from xrviz.dashboard import Dashboard
      4 

/opt/conda/lib/python3.6/site-packages/xrviz/__init__.py in <module>()
----> 1 from . import sample_data
      2 
      3 from ._version import get_versions
      4 __version__ = get_versions()['version']
      5 del get_versions

/opt/conda/lib/python3.6/site-packages/xrviz/sample_data/__init__.py in <module>()
      5         os.path.abspath(
      6             os.path.join(
----> 7                 os.path.dirname(__file__)+'/great_lakes.nc')))
```

Can we use the [pyviz approach to load the example data](http://pyviz.org/installation.html), something like running `xrviz examples`?

jsignell:
I think this can be closed.

PR #23, opened 2019-07-13
sample_data import for examples

gwbischof:
Fixes https://github.com/intake/xrviz/issues/5


martindurant:
@gwbischof , this functionality has moved into the package itself, under `xrviz.example`.

PR #24, opened 2019-07-16
Change holoviews version

hdsingh:
I have change added `holoviews 1.12.1` in requirements.txt

There is a minor problem that a pip user will face once we have `holoviews 1.12.1`:
Doing `python setup.py install` or `pip install xrviz` will show everything except projection panel.
To view projection, doing `conda install geoviews` will again upgrade `holoviews`, as a result of which correct lat/lon will not be displayed. 

To correctly display lat/lon, `pip install holoviews==1.12.1` would be required.
So both `conda install geoviews` and `pip install holoviews==1.12.1` should be done if installed xrviz via pip.

Everything works fine for conda, if default correct dependencies are present.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,5 @@
 panel
+holoviews==1.12.1
 hvplot
 xarray
 netcdf4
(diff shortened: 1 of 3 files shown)


hdsingh:
I also added a spacer between `Plot Dimensions` and `Aggregations` in fields panel, since this version of holoviews didn't do that by default.

martindurant:
`test_fields_initial` needs updating

martindurant:
Thanks. This will be in 0.1.0

PR #25, opened 2019-07-17
Change displayer width policy to min

hdsingh:
The variable displayer in `Variables panel` automatically kept on expanding if the variable is longer in length(ex most variables in HRRR data), since with policy was initially `max`. This has been changed to minimum.

### DIFF
--- a/xrviz/display.py
+++ b/xrviz/display.py
@@ -25,9 +25,9 @@ class Display(SigSlot):
     def __init__(self, data):
         super().__init__()
         self.data = data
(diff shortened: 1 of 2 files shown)


martindurant:
Can you show screenshots demonstrating this change, please?

hdsingh:
Earlier:
Screenshot from 2019-07-18 02-18-45

Now the width of displayer does not change automatically.
Screenshot from 2019-07-18 02-30-50

martindurant:
OK; although the option of a horizontal scroll bar in the selector box would have been nice too.

PR #26, opened 2019-07-17
Tests

hdsingh:
I have added more tests. `pytest --cov` gives 97%.

@martindurant In `projections` I have written tests using `parametrize`, while in `style` the tests are simple functions. Please let me know if you also feel that `parametrize` approach is better and tests in `style` should be changed according to it.


martindurant:
I like 97%!
Perhaps next week we can discuss how to hook into coveralls, so that this excellent coverage shows on the repo main screen as a badge of good code.

martindurant:
> Please let me know if you also feel that parametrize approach is better and tests in style should be changed according to it.

I have no strong opinion on this, it's like comparing writing out a statement three times to making a for look. One is more explicit, the other is more compact. I haven't looked at the tests themselves yet, but I'm unlikely to ask you to change just for aesthetics.

martindurant:
I would recommend at least a couple of tests which directly call SigSlot methods, rather than setting values on widgets.

hdsingh:
@martindurant I have added tests using `_emit` method. Can this PR be merged now?

PR #27, opened 2019-07-17
Series extraction

hdsingh:
Initially extraction was implemented for the case when both `x` and `y` are dims. 
Tasks Completed:
- [x] Shift `Extract Along` selector down.
- [x] Link `Extract Along` with `Clear` button.
- [x] Add `None` in options of `Extract Along`.
- [x] Implement extraction for the case when both of `x`,`y` are coords.
- [x] Hover to display value of extra dims

Screenshot from 2019-07-17 16-47-16


hdsingh:
Screenshot from 2019-07-20 09-58-58

@rsignell-usgs @martindurant Now we can extract for both coords and dims. Please have a look.

martindurant:
(note the conflict, above, since the merger of #26)

hdsingh:
Screenshot from 2019-07-21 00-00-36
I have also added custom hovers with extra information as discussed. 
@rsignell-usgs @martindurant This PR can be reviewed now.

rsignell-usgs:
I'm having trouble testing this.    I tried to follow the install instructions to create the environment:
```
conda install -c conda-forge xrviz
pip install git+https://github.com/intake/xrviz --no-deps
```
but then had to do:
```
conda install -c holoviews=1.12.1 datashader hvplot
```
before I could run the example:
```
```.

When I run the example, clicking on the map doesn't give any indication that anything has happened, and I don't get a time series plot.   

Does something need to be enabled/selected on the GUI first?

**Comment by hdsingh** [comment]
As discussed in previous meeting I have added `None` in `extract along`. So we need to select the dimension along which to extract in the fields.

**Comment by hdsingh** [comment]
> but then had to do:
`conda install -c holoviews=1.12.1 datashader hvplot`

I don't understand why there is need to this since these dependencies are already specified in [requirements.txt](https://github.com/intake/xrviz/blob/master/requirements.txt)

**Comment by martindurant** [comment]
It is better to get the deps from conda (the pip command was run with `--no-deps`)

**Comment by rsignell-usgs** [comment]
How about providing an `environment.yml` file so it's easy for conda users to know what to do?

**Comment by hdsingh** [comment]
Updating `conda xrviz` (https://github.com/conda-forge/xrviz-feedstock) will solve this. I will create a PR.

**Comment by martindurant** [comment]
^ was done, package should be built by now. @rsignell-usgs , do you want to try again?

**Comment by rsignell-usgs** [comment]
So try this again?
```
conda install -c conda-forge xrviz
pip install git+https://github.com/intake/xrviz --no-deps
```

**Comment by rsignell-usgs** [comment]
I did: 
```
(xrviz) C:\Users\rsignell\Documents\GitHub\xrviz>git remote -v
hdsingh https://github.com/hdsingh/xrviz.git (fetch)
hdsingh https://github.com/hdsingh/xrviz.git (push)
origin  https://github.com/intake/xrviz.git (fetch)
origin  https://github.com/intake/xrviz.git (push)

conda create -n xrviz -c conda-forge xrviz
conda activate xrviz
git fetch hdsingh
git checkout remotes/hdsingh/series_extraction
pip install -e xrviz
```
but I still don't get the time series panel popping up.

martindurant:
pip install -e .
?

On July 22, 2019 2:35:23 PM EDT, Rich Signell <notifications@github.com> wrote:
>I did: 
>```
>(xrviz) C:\Users\rsignell\Documents\GitHub\xrviz>git remote -v
>hdsingh https://github.com/hdsingh/xrviz.git (fetch)
>hdsingh https://github.com/hdsingh/xrviz.git (push)
>origin  https://github.com/intake/xrviz.git (fetch)
>origin  https://github.com/intake/xrviz.git (push)
>
>conda create -n xrviz -c conda-forge xrviz
>conda activate xrviz
>git fetch hdsingh
>git checkout remotes/hdsingh/series_extraction
>pip install -e xrviz
>python -c "import xrviz; xrviz.example()"
>```
>but I still don't get the time series panel popping up.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/intake/xrviz/pull/27#issuecomment-513902847

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

rsignell-usgs:
I think it would be a better user experience to have the default for "extract along" to be the first dimension e.g. "time" instead of "None".

martindurant:
Yet another example, where we are trying to guess what our users' expectation may be, which depend on whether they have used other multi-dim viz apps or not, whether they are climate people...

rsignell-usgs:
@martindurant , I think I was unclear what I meant (again).  I didn't mean to suggest choosing "time".  I just meant that whatever the first dimension is would be selected.  I think this would help users because they could click and get something appearing without clicking a dimension first.  And then select a different dimension if that's not what they wanted.

martindurant:
Not certain that a user would necessarily expect to have plots popping up, though. You may be right that that is reasonable to suppose, I am just raising a little caution because I'm not certain. I feel like I/we have already made a lot of decisions for people! Maybe it all comes down to the writing of good docs and tests, which I suspect will take up the majority of the remaining time.

hdsingh:
@rsignell-usgs @martindurant According to me work here is complete. I have also added tests. Can you please review this PR so that it could be merged well in time?

rsignell-usgs:
Okay, let's merge this.

martindurant:
Excellent!

PR #28, opened 2019-07-17
Urgent fixes

hdsingh:
I have implemented the following:
1. Move `rasterize` to Style Panel
2. Use TextInput for `crs_params` and `proj_params`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ matrix:
         # Install dependencies
         - conda create -n test python=3.7 pytest
         - source activate test
-        - conda install xarray cartopy panel hvplot netcdf4
+        - conda install xarray cartopy panel hvplot netcdf4 datashader
 
       script:
         - pip install . --no-deps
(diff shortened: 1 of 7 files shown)


martindurant:
> Use TextInput for crs_params and proj_params

Why this? You immediately do literal_eval anyway.

I meant that, instead of including `{'projection': crs.thing}` in your box, you can have `{'projection': 'thing'}`.

hdsingh:
It seems from the failed test that we need to add `datashader` as a dependency. Earlier it was installed with `geoviews`. Or shall we make it optional like `metpy`?

martindurant:
If rasterize is normally True, we need to include datashader explicitly. Otherwise, we would need to check for it and disable that option when it isn't present. I say add the dependency.

hdsingh:
Now we can have 
```python
init_params = {'is_geo': True,
               'basemap': 'OSM',
               'crs params': "{'central_longitude': 0,  'globe': {'ellipse':'sphere'}}"}       
```

martindurant:
Yes, that's exactly how I think it should be. Merging this now (urgently!) to aid creation of content for presentation.

Issue #30, opened 2019-08-02
Top level callable


martindurant:
I recommend that there should be something in the top-level `__init__` which is the main entry-point for users, e.g., 
```
from .dashboard import Dashboard as XRViz
```

so that we can recommend the normal usage as
```
import xrviz
xrviz.XRViz(data, ...).show()
```

PR #32, opened 2019-08-03
Improve docs

hdsingh:
I will improve documentation in this PR.


xrviz/__init__.py
```diff
@@ -1,3 +1,4 @@
+from .dashboard import Dashboard as XrViz
```
hdsingh: Since only 1 line change is required, I have added it here.
Now we can do
```
import xrviz
xrviz.XrViz(data, ...).show()
```
as stated in https://github.com/intake/xrviz/issues/30. 
(replaced `XRViz` with `XrViz` to keep it consistent everywhere)

hdsingh:
I have also added `sphinx_rtd_theme` (that is being used by intake docs).

hdsingh:
@martindurant The API doc looks good to me now. Please have a look.

xrviz/control.py
```diff
@@ -11,24 +11,29 @@
...
-    such as displayer, fields.
+    This section arranges all the panes in the form of tabs,
+    hence providing the controls to inteact with all the
+    panes.
 
     Parameters
     ----------
-    data: `xarray` instance: `DataSet` or `DataArray`
-           datset is used to initialize.
+    data: `xr.core.dataarray.DataWithCoords`
```
martindurant: Prefer previous, simple description, `xarray.DataSet` or `xarray.DataArray`
You can then specify in the description what that means. I don't expect users to know what DataWithCoords is.

xrviz/control.py
```diff
@@ -11,24 +11,29 @@
...
+        Data is required for initialization.
```
martindurant: This is obvious from the signature, and that you haven't said (optional) or given a default. Instead, we want to say what the role of this input is.

xrviz/control.py
```diff
@@ -11,24 +11,29 @@
...
 
     Attributes
     ----------
-    panel: Displays the generated template.
-    displayer: Provides access to `Display` sub-section.
-    describer: Provides access to `Describe` sub-section.
-    fields: Provides access to `Fields` sub-section.
-    kwargs: Provides access to kwargs selected in different subsections.
+    1. ``panel``: Displays all the panes arraged in form of tabs.
```
martindurant: A `panel.Tabs` instance containing the user input panes of the interface.

xrviz/control.py
```diff
@@ -11,24 +11,29 @@
...
+    2. ``displayer``: Provides access to `Display` sub-section of `Variables` pane.
```
martindurant: This is not a useful description.

"A `SigSlot` instance which shows..."

xrviz/control.py
```diff
@@ -11,24 +11,29 @@
...
+    3. ``describer``: Provides access to `Describe` sub-section of `Variables` pane.
+    4. ``coord_setter``: Provides access to the `Set Coords` pane.
+    5. ``fields``: Provides access to `Axes` pane.
+    6. ``style``: Provides access to `Style` pane.
+    7. ``projection``: Provides access to `Projection` pane (if present).
+    8. ``kwargs``: Provides access to values selected in different panes.
```
martindurant: None of these "provides access" statements are very useful. An attribute always provides access to something. Be very specific, as a scientist would. This last one, for instance:
"a dictionary gathered from the widgets of the input Panes, of a form which can be passed to the plotting function as kwargs"

xrviz/control.py
```diff
@@ -11,24 +11,29 @@
...
     """
 
     def __init__(self, data):
+        """Initializes the Control pane."""
```
martindurant: This is not useful text. Initialising is what an `__init__` method always does. You could leave an empty string, or again describe what `data` is.

xrviz/control.py
```diff
@@ -66,7 +71,7 @@ def setup_initial_values(self, initial_params={}):
 
     def set_coords(self, data):
         """
-        To set the data coords
+        To set the data coordinates.
```
martindurant: Actually this method does not set coordinates at all, right?
Instead, it is called after coordinates have been set, to update the other input panes.

It is generally better to use descriptive statements "apply data with new coordinate sets to all input panes"

xrviz/coord_setter.py
```diff
@@ -3,10 +3,20 @@
 
 
 class CoordSetter(SigSlot):
+    """
+    To set and reset the data coordinates.
```
martindurant: An input pane for choosing which variables are considered coordinates

xrviz/coord_setter.py
```diff
@@ -3,10 +3,20 @@
...
 
+    It uses a `Cross Selector <https://panel.pyviz.org/reference/widgets/CrossSelector.html>`_
+    to display a list of simple and coordinate variables.
+    Simple variables (which are not data coordinates) are available on
+    left side and default coordinates are available on right side.
+    To set variables as coordinates, make selection on left side and click
+    ``>>``. Similarly making selection on right side and clicking ``<<``
+    will reset the coordinates. Other panes would update themselves
```
martindurant: remove "would" - this is not conditional, it happens every time

xrviz/coord_setter.py
```diff
@@ -3,10 +3,20 @@
...
+    accordingly, in response to this change.
+    """
     def __init__(self, data):
-        """
-        To intialize
+        """ Initializes the CoordSetter object.
```
martindurant: Not useful

martindurant: Also, as a matter of definition, this initialises an **instance**. The word "object" is ambiguous in python, and you should not use it unless you are referring to `object` (which is an instance of `type`, i.e., a class).

xrviz/coord_setter.py
```diff
@@ -18,7 +28,7 @@ def __init__(self, data):
 
     def set_coords(self, data):
         """
-        to set the coords
+        To set the coords
```
martindurant: Called when the data attribute of the interface acquires new coordinates

Why is this ever called at all? In this pane, we *define* the coordinates, so this may lead to recursion.

xrviz/dashboard.py
```diff
@@ -20,23 +20,28 @@
 class Dashboard(SigSlot):
     """
     This section provides access to the complete generated dashboard,
```
martindurant: Since this is object-oriented programming, this **is** the dashboard. It is not a section, and it does not provide access, it simply is.

"This is the main entry point to XrViz, and generates an interactive GUI for a given dataset"

xrviz/dashboard.py
```diff
@@ -20,23 +20,28 @@
...
-    consisting of all subsections.
+    consisting of all the panes.
 
     Parameters
     ----------
-    data: `xarray` instance: `DataSet` or `DataArray`
-           datset is used to initialize.
-    initial_params: To pre-set values of widgets.
+    data: `xr.core.dataarray.DataWithCoords`
```
martindurant: As above. I believe the other cases take only `xarray.DataSet`. Note that you should not abbreviate with `xr`.

xrviz/dashboard.py
```diff
@@ -20,23 +20,28 @@
...
+        Data is required for initialization.
+    initial_params: `dict`
+        To pre-select values of widgets upon initialization.
```
martindurant: Where can we find out about the keys and values allowed here?

xrviz/dashboard.py
```diff
@@ -20,23 +20,28 @@
...
 
     Attributes
     ----------
-    panel: Displays the generated dashboard.
-    control: Provides access to the generated control panel.
-    plot: Plot button, upon click generates graph according to
-          kwargs selected in other sub-sections.
-    output: Provides access to generated graph.
+    1. ``panel``: Displays the generated dashboard.
```
martindurant: Same comments as above

xrviz/dashboard.py
```diff
@@ -20,23 +20,28 @@
...
+    2. ``control``: Provides access to the generated control panel.
+    3. ``plot_button``: Upon click generates graph according to kwargs selected in other sub-sections.
+    4. ``graph``: Provides access to the main graph object.
+    5. ``output``: Provides access to generated graph object with the index selectors.
+    6. ``taps_graph``: Provides access to the graph having location of taps.
+    7. ``series_graph``: Provides access to the graph having series extracted.
+    8. ``clear_series_button``: Button to clear the `taps_graph` and `series_graph`.
     """
     def __init__(self, data, initial_params={}):
+        """Initializes the Dashboard."""
```
martindurant: Not useful

xrviz/dashboard.py
```diff
@@ -97,16 +102,19 @@ def clear_series(self, *args):
 
     def link_aggregation_selectors(self, *args):
         """
-        To link aggregation selectors with cmap limits
```
martindurant: This is an internal method, could prepend `_` to the name and maybe remove docstring

xrviz/dashboard.py
```diff
@@ -97,16 +102,19 @@ def clear_series(self, *args):
...
         for dim_selector in self.control.kwargs['remaining_dims']:
             self.control.fields.connect(dim_selector, self.control.style.setup)
 
     def create_plot(self, *args):
         """
-        To create plot.
+        This method creates a plot according to the values selected in the
+        widgets. It handles the following two cases:
+            1. Both `x`, `y` are present in selected variable's coordinates.
+            2. One or both of  `x`, `y` are NOT present in selected variable's coordinates.
```
martindurant: Please elaborate

xrviz/dashboard.py
```diff
@@ -240,7 +248,8 @@ def create_plot(self, *args):
 
     def create_indexed_graph(self, *args):
         """
-        Creates graph for  selected indexes in selectors or players.
+        Creates a graph for the dimensions selected in `x` and `y`. This
```
martindurant: What is a graph, opposed to a plot?

xrviz/dashboard.py
```diff
@@ -308,7 +317,10 @@ def create_indexed_graph(self, *args):
 
     def create_taps_graph(self, x, y, clear=False):
         """
-        To mark taps
+        This method return a graph to record the taps, when
```
martindurant: "Create an output layer in the graph which responds to taps from the user

Whenever the user taps (or clicks) the graph, a glyph will be overlaid, and a series extracted at that point"

xrviz/dashboard.py
```diff
@@ -330,14 +342,24 @@ def create_taps_graph(self, x, y, clear=False):
 
     def create_series_graph(self, x, y, color, clear=False):
         """
-        Create series graph
+        This method extracts a series at the point where user has tapped
```
martindurant: Extract a series at a given point, and plot it

xrviz/dashboard.py
```diff
@@ -457,7 +478,8 @@ def set_coords(self, *args):
 
     def check_is_plottable(self, var):
         """
-        If a variable is 1-d, disable plot_button for it.
+        Check if a data variable can be plotted or not. If a variable is 1-d,
```
martindurant: Remove "or not"

xrviz/coord_setter.py
hdsingh: The actual set/reset of coordinates happens in `dashboard.py`. Here `self.coord_selector.value = list(self.data.coords)` ensures that the default coords are also present by interpreting from the data, in case they  have been reset. Here just updated values are being displayed.

martindurant: I disagree: at `self.coord_selector = ` in `__init__`, you set the starting values. Now whenever `self.coord_selector.value` changes, because of user action, the outer class takes care to set the coordinates on the xarray object, but why do you need to set the `.value` again? I believe this will always be a no-op (but you could set a breakpoint or log here, if you don't think so).

hdsingh: I have tried removing `self.coord_selector.value = list(self.data.coords)`. If the user removes `time` or `sigma`, these(indexed coord) would not be available in the values in `Set Coords` but would be present as coords in data and will be displayed as coords in `Axes`. This would be inconsistent.

> why do you need to set the `.value` again?

To get back the values of indexed coordinates in case they have been reset by the user. (The user can select and try resetting an indexed coord but it would again be added to the values on right with the help of this line).

I have added `logger.debug(self.coord_selector.value)` before setting `.value`.

hdsingh: > I believe this will always be a no-op.

Yes, it is no-op but only for the case when the value that has been set/reset is not indexed coord.

xrviz/dashboard.py
hdsingh: I don't think there is any difference. I will replace `create_plot` with `create_graph` to have similar function names.

martindurant:
If you haven't already, please include `"numpydoc"` in the extensions listed in conf.py and add it in docs/environment.yml

xrviz/dashboard.py
hdsingh: Added link to `Set Initial Parameters` doc.

hdsingh:
@martindurant Thanks a lot for such detailed reviews! These are really helpful for me to know more and understand the way docs should be written.

I have made changes according to the review. Please have a look.

martindurant:
I went through a lot of the docstrings trying to make the text more concise. In the meantime, I may have applied some style guidelines to the code... Please review what I did in the last commit, and apply similar patterns elsewhere. 

I still have reservations about exactly what happens when coordinate widgets are updated, I'll chase that up after this PR. I am finding that the xrviz interface within the intake interface is amazingly slow to set widget values, while dfviz is not, and it is essential to get to the bottom of why that might be.

hdsingh:
@martindurant I made similar changes and reformat code where required. Documentation looks much better now.
According to me work here is complete. Please let me know if something else needs to be changed  in order to merge this PR.

martindurant:
I think we may still want to circle around to this again, but I'll merge it now, as the PR has become rather long.

PR #37, opened 2019-08-09
Handle dask arrays

hdsingh:
Now we can create plot for dask arrays.

```python
import xarray as xr
from xrviz.dashboard import Dashboard

ds = xr.tutorial.open_dataset('air_temperature',
                              chunks={'lat': 25, 'lon': 25, 'time': -1})
dash = Dashboard(ds)
dash.show()
```
The above code now runs without error (on clicking `PLOT`). 

In master branch it gives `TypeError: quantile does not work for arrays stored as dask arrays. Load the data via .compute() or .load() prior to calling this method.`

### DIFF
--- a/xrviz/dashboard.py
+++ b/xrviz/dashboard.py
@@ -296,6 +296,7 @@ class Dashboard(SigSlot):
         This is used when values selected in `x` and `y` are not data
         coordinates (i.e. one or both values are data dimensions).
         """
+        self.kwargs = self.control.kwargs
         selection = {}  # to collect the value of index selectors
         for i, dim in enumerate(list(self.var_selector_dims)):
             selection[dim] = self.index_selectors[i].value
--- a/xrviz/dashboard.py
+++ b/xrviz/dashboard.py
@@ -188,7 +188,7 @@ class Dashboard(SigSlot):
 
             # It is better to set initial values as 0.1,0.9 rather than 0,1(min, max)
             # to get a color balance graph
-            c_lim_lower, c_lim_upper = (float(cmin), float(cmax)) if cmin and cmax else ([q for q in sel_data_for_cmap.quantile([0.1, 0.9])])
+            c_lim_lower, c_lim_upper = (float(cmin), float(cmax)) if cmin and cmax else ([q for q in sel_data_for_cmap.compute().quantile([0.1, 0.9])])
 
             color_range = {sel_data.name: (c_lim_lower, c_lim_upper)}
 
@@ -283,7 +283,7 @@ class Dashboard(SigSlot):
 
         # It is better to set initial values as 0.1,0.9 rather than 0,1(min, max)
         # to get a color balance graph
-        c_lim_lower, c_lim_upper = (float(cmin), float(cmax)) if cmin and cmax else ([q for q in sel_data.quantile([0.1, 0.9])])
+        c_lim_lower, c_lim_upper = (float(cmin), float(cmax)) if cmin and cmax else ([q for q in sel_data.compute().quantile([0.1, 0.9])])
 
         color_range = {sel_data.name: (c_lim_lower, c_lim_upper)}
 
--- a/xrviz/dashboard.py
+++ b/xrviz/dashboard.py
@@ -15,7 +15,7 @@ import numpy
 from .sigslot import SigSlot
 from .control import Control
 from .utils import convert_widget, player_with_name_and_value, is_float
-from .compatibility import ccrs, gv, gf, has_cartopy, logger
+from .compatibility import ccrs, gv, gf, has_cartopy, logger, has_crick
 
 
 class Dashboard(SigSlot):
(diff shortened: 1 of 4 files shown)


martindurant:
This is a terrible idea!
The whole idea of a dask-based xarray is that it is too big to fit into memory, and maybe needs to be processed on a cluster. `.compute()` brings the whole thing into memory. So it will be ok for the case that you are using only one slice, but **not** for the whole array. For the latter case at least, I would certainly use min/max instead of percentiles.

(note that the code should be tested anyway - I'm actually not sure whether `.compute()`, the dask method, is the right thing here, instead of xarray's `.values`).

hdsingh:
I will go through the dask docs before attempting to solve this again, to get better understanding.

martindurant:
OK, but ask questions sooner rather than later - dask is a pretty big project, and you are looking specifically at the xarray interface, which hides/augments some of the dask.array functionality.

xrviz/dashboard.py
```diff
@@ -576,6 +575,13 @@ def both_coords_2d_with_same_dims(self):
         return len(x_dims) == len(y_dims) == 2 and sorted(x_dims) == sorted(y_dims)
 
 
+def find_cmap_limits(sel_data):
+    if isinstance(sel_data.data, dask.array.core.Array):
+        return dask.array.percentile(sel_data.data.ravel(), (10, 90)).compute()
+    else:  # if sel_data.data is numpy.ndarray
+        return [float(q) for q in sel_data.quantile([0.1, 0.9])]
+
```
hdsingh: @martindurant What do you think about this approach?

hdsingh:
Use of `method=tdigest` in `dask.array.percentile` would require `crick` and `cython` as dependency. Shall it be used?

xrviz/dashboard.py
```diff
@@ -576,6 +576,13 @@ def both_coords_2d_with_same_dims(self):
         return len(x_dims) == len(y_dims) == 2 and sorted(x_dims) == sorted(y_dims)
 
 
+def find_cmap_limits(sel_data):
+    if isinstance(sel_data.data, dask.array.core.Array):
+        return dask.array.percentile(sel_data.data.ravel(), (10, 90)).compute()
+    else:  # if sel_data.data is numpy.ndarray
```
martindurant: Note that in the future, this could also be other things like cupy arrays

martindurant:
OK, we can do this for now. Maybe it will change later.

Note that we can test for the existence of crick.TDIgest and use it, if possible. We should not require it. Also, crick does *not* depend on cython, only numpy ( https://github.com/conda-forge/crick-feedstock/blob/master/recipe/meta.yaml#L26 ). Cython is used during building within conda-forge (or pip, if you install from source). If you do not know, I can give a brief introduction into what cython does at our next meeting.

Please test and use crick if possible, and then we can merge this.

I would like you at some point to justify the rounding to 5 places.

xrviz/dashboard.py
```diff
@@ -339,14 +340,13 @@ def create_indexed_graph(self, *args):
...
-            else ([q for q in sel_data.quantile([0.1, 0.9])])
-        )
+            else find_cmap_limits(sel_data))
 
         color_range = {sel_data.name: (c_lim_lower, c_lim_upper)}
 
         if not cmin:  # if user left blank or initial values are empty
-            self.control.style.lower_limit.value = str(c_lim_lower.values.round(5))
-            self.control.style.upper_limit.value = str(c_lim_upper.values.round(5))
+            self.control.style.lower_limit.value = str(round(c_lim_lower, 5))
```
hdsingh: Without rounding, the float values can be upto 16 decimal places (`5.454999923706055`) which might not look good, so just for presentation purposes I am rounding off to 5.
This is not necessary and can be removed.

martindurant: It may be right, but we should own that it is an arbitrary decision.

hdsingh: I will mention about this in docs.

hdsingh:
I have made relevant changes. Please have a look.

xrviz/tests/test_dashboard.py
```diff
@@ -211,3 +211,11 @@ def test_create_taps_and_series_graph_for_2d_coords(dashboard):
...
     dashboard.create_taps_graph(x=-79.232, y=43.273)
     assert isinstance(dashboard.series_graph[0], pn.pane.holoviews.HoloViews)
+
+
+@pytest.mark.skipif(not has_crick_tdigest, reason='crick.tdigest not present')
+def test_find_cmap_limits_with_crick_tdigest():
+    ds = xr.tutorial.open_dataset('air_temperature',
+                                  chunks={'lat': 25, 'lon': 25, 'time': 10})
+    a, b = find_cmap_limits(ds.air)
+    assert (a, b) == (255.38780056044027, 298.5900340551101)
```
martindurant: These numbers should not the exact like this, precision may differ by system. Probably it is good enough if `abs(expected - found) < 0.1`

hdsingh: Corrected. Thanks!

martindurant:
OK, going in when it turns green.

PR #49, opened 2019-10-15
Fixing selector not showing up, taps not working in geo

jsignell:
Fixing some broken functionality that I saw with the latest releases.

### DIFF
--- a/xrviz/dashboard.py
+++ b/xrviz/dashboard.py
@@ -14,7 +14,7 @@ from itertools import cycle
 import numpy
 from .sigslot import SigSlot
 from .control import Control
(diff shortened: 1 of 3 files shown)


martindurant:
I trust you on this, you don't need my say-so.
@hdsingh any comments? Will merge in 24hr if not.

hdsingh:
Thanks a lot for this PR! I am very happy to see that XrViz is being looked after!

This `look_for_class` method is so much better than the naive `try/except` approach.

jsignell:
Yeah the main issue was that the nesting changed with more recent panel, so the widgets were just silently failing to show up.

PR #51, opened 2019-10-25
Using new xarray html repr for describe

jsignell:
This is for when xarray 0.14.1 is released (which will be soon). 

Screen Shot 2019-10-25 at 1 32 38 PM


jsignell:
I'm still thinking about how to make the repr itself something that the user can select on. So that there doesn't have to be duplicated lists.

martindurant:
If showing the repr, then the choice really can be from the simpler combo box, which only shows one clickable label.
Will this break your vertical space requirements?

jsignell:
> If showing the repr, then the choice really can be from the simpler combo box, which only shows one clickable label.

Or it could even be an autocomplete input. Which feels even more usable to me.

> Will this break your vertical space requirements?

I gave the repr a max_height and scroll on overflow, so it'll never get too big.

martindurant:
> it could even be an autocomplete input 

Would love to see it, if you know how.

jsignell:
Unfortunately it wasn't exposed yet to allow autocompletion even for 1 character. I opened the PR on panel. https://github.com/pyviz/panel/pull/721

jsignell:
Ok now that xarray and panel have been released, I'm going to clean up any tests and merge this PR.

martindurant:
hurray!

PR #52, opened 2019-10-25
Adding help to tabs

jsignell:
As I've been messing around with xrviz I've found that some of the tabs could really do with some minimal help text. It a challenge to include enough to get people stared, but not so much that it takes space away from the real work. Anyways, just wanted to open this PR so I don't forget that I started it. 

I was also wondering if there should be a help tab that would have knowledge of the last tab that you were on and provide more detailed help relevant to the actions on that tab.

### DIFF
--- a/xrviz/control.py
+++ b/xrviz/control.py
@@ -8,6 +8,11 @@ from .style import Style
 from .coord_setter import CoordSetter
 from .compatibility import has_cartopy
 
+TEXT = """
+Select a data variable to visualize. For more information,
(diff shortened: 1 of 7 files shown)


martindurant:
As a separate tab seems odd - but we do have the floating buttons, so the plot area could be filled by help text on whatever is currently selected.

jsignell:
Yeah maybe links to the docs from text in the tab would do the trick.

martindurant:
Works for me

PR #59, opened 2019-11-26
Switch autocomplete back to dropdown

### LINKED ISSUE #57: Return to dropdown selection selection of variables
I think the "start typing w/ autocomplete" is cool, but I think it's more intuitive (and even easier) to pick a variable from a drop down list, as we had before.   A new user (including me a few minutes ago) wouldn't know what to do when encountering this:

jsignell:
I had switched the variable input to autocomplete, and here I am switching to select. 

Before it was autocomplete, the input was a multiselect which I don't think  was appropriate or easier to use.

Closes #57


jsignell:
Screen Shot 2019-11-26 at 12 19 16 PM
Screen Shot 2019-11-26 at 12 19 34 PM

PR #69, opened 2020-04-17
Check metpy version

hdsingh:
Since `metpy 1.0` uses 
```
x, y = parsed_var.metpy.longitude, parsed_var.metpy.latitude 
```

instead of 
```
x, y = parsed_var.metpy.coordinates('x', 'y')
``` 
as in previous versions, a version check has been added.

### DIFF
--- a/xrviz/fields.py
+++ b/xrviz/fields.py
@@ -1,4 +1,3 @@
-import metpy
 import panel as pn
 from packaging.version import Version
 import xarray as xr
@@ -243,6 +242,7 @@ class Fields(SigSlot):
         .. _`metpy.parse_cf`: https://github.com/Unidata/MetPy/blob/master/metpy/xarray.py#L335
         """
         try:
+            import metpy
             parsed_var = self.data.metpy.parse_cf(var)
             if(Version(metpy.__version__)) > Version("0.12"):
                 x, y = parsed_var.metpy.longitude, parsed_var.metpy.latitude 
--- a/xrviz/fields.py
+++ b/xrviz/fields.py
@@ -1,4 +1,6 @@
+import metpy
 import panel as pn
+from packaging.version import Version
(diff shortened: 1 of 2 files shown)


hdsingh:
@martindurant  Tests are failing only in case of `Python 3.7`, looks like due to some change in `netCDF4`.

xrviz/fields.py
```diff
@@ -1,4 +1,5 @@
 import panel as pn
+from packaging.version import Version
```
martindurant, line 2: You might want `LooseVersion`, because the version string could contain non-standard parts such as "rc", "beta", or git information.

xrviz/fields.py
```diff
@@ -241,8 +242,12 @@ def guess_x_y(self, var):
         .. _`metpy.parse_cf`: https://github.com/Unidata/MetPy/blob/master/metpy/xarray.py#L335
         """
         try:
+            import metpy
             parsed_var = self.data.metpy.parse_cf(var)
-            x, y = parsed_var.metpy.coordinates('x', 'y')
+            if(Version(metpy.__version__)) > Version("0.12"):
+                x, y = parsed_var.metpy.longitude, parsed_var.metpy.latitude 
+            else:
+                x, y = parsed_var.metpy.coordinates('x', 'y')
```
martindurant, line 250: So this doesn't work for 0.12, but it still gives a warning? In that case, we should explicitly suppress the warning here. 
```
with warnings.catch_warnings():
    warnings.filterwarnings("ignore", category=DeprecationWarning)
    x, y, = ...
```

martindurant:
The cython error means that there is a mismatch between the versions of python and/or cython used to build the various binaries installed by pip - naughty pip for not understanding requirements! Do you think you can play with some version pinning to get it running? This is not really our responsibiilty...

